PR #7, opened 2018-03-17
Add deployment endpoints

orasimus:


valohai_yaml/objs/endpoint.py
```diff
@@ -0,0 +1,42 @@
...
+
+        assert all(isinstance(f, File) for f in files)
+        self.files = files
+
+    @classmethod
+    def parse(cls, data):
+        kwargs = data.copy()
+        kwargs['files'] = [File.parse(f_data) for f_data in kwargs.pop('files', ())]
+        if 'server-command' in kwargs:
+            kwargs['server_command'] = kwargs.pop('server-command')
```
orasimus: Is there a better way to do this, i.e. parse the dash as an underscore?

valohai_yaml/objs/config.py
```diff
@@ -62,7 +72,9 @@ def get_step_by(self, **kwargs):
         return None
 
     def __repr__(self):  # pragma: no cover
-        return '<Config with %d steps: %r>' % (
+        return '<Config with %d steps and %d endpoints: %r, %r>' % (
```
akx: This would probably be clearer if the numbers and names were paired up, like

> Config with %d steps (%r) and %d endpoints (%r)

?

valohai_yaml/objs/endpoint.py
akx: `_SimpleObject`'s `parse` and `serialize` methods do this roundtripping.

valohai_yaml/schema/endpoint.yaml
```diff
@@ -0,0 +1,49 @@
+---
+oneOf:
+- type: object
+  required:
+  - name
+  - image
+  - port
```
akx: `port` is not `required`; it defaults to 8000.

valohai_yaml/schema/endpoint.yaml
```diff
@@ -0,0 +1,49 @@
+---
+oneOf:
```
akx: I think we shouldn't need an `oneOf` structure to express that either `server-command` or `wsgi` (or future variants) is required.

Instead, add the `oneOf` validation structure within the object, as discussed here: https://stackoverflow.com/a/25637397/51685

akx: Also, it would be quite lovely if all of the fields here also had a good `description:` stanza, so as to make the format more self-documenting.

valohai_yaml/schema/file-item.yaml
```diff
@@ -0,0 +1,16 @@
...
+properties:
+  name:
+    type: string
+    description: The symbolic name of this file.
+  description:
+    type: string
+    description: The description of the file.
+  path:
+    type: string
+    description: The path of the file.
```
akx: This is actually the relative path to the file from the endpoint's working directory. An example or two in the `description` might help, e.g. just adding "e.g. `data/serialized_model.hdf5`".

tests/test_parsing.py
```diff
@@ -83,3 +84,15 @@ def test_boolean_param_parse(boolean_param_config):
...
+def test_endpoint_parse(endpoint_config):
+    server_endpoint = endpoint_config.endpoints['server-endpoint']
+    assert server_endpoint.image == 'python:3.6'
+    assert server_endpoint.port == 1453
+    assert server_endpoint.server_command == 'python run_server.py'
+    wsgi_endpoint = endpoint_config.endpoints['wsgi-endpoint']
+    assert wsgi_endpoint.description == 'predict digits from image inputs'
+    assert wsgi_endpoint.image == 'gcr.io/tensorflow/tensorflow:1.3.0-py3'
+    assert wsgi_endpoint.wsgi == 'predict_wsgi:predict_wsgi'
+    assert len(wsgi_endpoint.files) == 1
```
akx: The file object's content is not tested though?

akx: Also, mind adding a test for serialization/roundtripping of endpoint data?

valohai_yaml/objs/endpoint.py
akx: So yeah, please just inherit from `_SimpleObject` and override parts of the `get_data()` method (mostly `files`).

valohai_yaml/objs/config.py
orasimus: 👍

tests/test_parsing.py
orasimus: The tests are already doing this, they are running each test twice, once with and once without doing the roundtripping.

orasimus: Now they are :)

PR #12, opened 2019-01-28
Schema additions

akx:
* multiple default URLs
* default environment variables
* default execution environment


examples/example3.yaml
```diff
@@ -5,6 +5,9 @@
     image: busybox
     command:
      - ls -laR /valohai/inputs
+    environment-variables:
+      foo: bar
+      baz: 850
```
ruksi: should we use the same syntax with environment variables as with parameters and inputs

```
environment-variables:
  - name: foo
    default: bar
```

I can imagine situations where we want to specify that "AWS_SECRET_ID" is required but is not provided in the configuration? 

Also, do environment variables require typing? They are all strings but if we want to "hyperparameter optimize" ENV VAR related tasks, they need a type.

ruksi: Then you can also add `description` to each ENV VAR, like explain what that arcane "IMGEN_LOOP" means ;D

akx: I guess we could use that syntax. It mirrors that of k8s's then. 

https://github.com/kubernetes/website/blob/d83bcfe033fb71cc1eb87eb17980852c2cb5a208/content/en/examples/pods/inject/envars.yaml#L12-L15

Adding `required: true` (or `optional: false`, to be coherent) is also a good idea.

Typing is not required here, as doing hyperparameter optimization on env vars would be done via parameters that are passed as environment variables (which isn't yet possible).

akx:
@ruksi Ready for rereview :)

PR #13, opened 2019-02-14
Fix flag parsing and move in linter from cli

JuhaKiili:
- Flags set to `default: true` will not appear when undefined
- New tests to validate the behavior of `optional` & `default` for all types
- Linting moved from `valohai-cli` to `valohai-yaml`
- Warning for setting `optional` property for type `'flag'`


valohai_yaml/__main__.py
```diff
@@ -4,31 +4,47 @@
...
 from collections import defaultdict
 
+from valohai_yaml.lint import lint
 from .validation import validate
 
 
 def main(args=None):
     ap = argparse.ArgumentParser()
     ap.add_argument('file', nargs='+', help='file(s) to validate')
+    ap.add_argument('--lint', action="store_true", help='Use linter')
```
akx: I don't think this needs to be an option – the linter is a superset of the validator, so we can just as well always use it.

valohai_yaml/lint.py
```diff
@@ -0,0 +1,87 @@
...
+            if el not in ('properties', 'items')
+        ]
+        obj_path = [str(el) for el in error.path]
+        lr.add_error('  {validator} validation on {schema_path}: {message} ({path})'.format(
+            validator=style(error.validator.title(), bold=True),
+            schema_path=style('.'.join(simplified_schema_path), bold=True),
+            message=style(error.message, fg='red'),
+            path=style('.'.join(obj_path), bold=True),
+        ))
+
```
akx: I forgot to mention this in meatspace, but we need to stop if there are errors at this point; we can't `Config.parse()` a schema-invalid document and expect good things.

JuhaKiili: That one failed test actually made me realize this. Fix pushed.

valohai_yaml/__main__.py
JuhaKiili: Fix pushed

valohai_yaml/lint.py
```diff
@@ -0,0 +1,90 @@
...
+    if len(errors) > 0:
```
akx: `if errors:` is equivalent and faster.

PR #27, opened 2020-03-02
Move valohai.yaml related things from valohai-utils to valohai-yaml

JuhaKiili:


valohai_yaml/utils/merge.py
```diff
@@ -0,0 +1,57 @@
...
+    return out
+
+
+def _merge_simple(a, b):
+    a = copy.deepcopy(a)
+    a.__dict__.update(copy.deepcopy(b).__dict__)
+    return a
+
+
+def _merge_config(a: Config, b: Config) -> Config:
```
akx: I think these should now be in the objects themselves - augment the `Item` base class to have a not-implemented merging method, then implement it in the subclasses where it's relevant:

```python
class Item:
    def merge_with(self, other) -> 'Item':
        raise NotImplemented('...')

class Config(Item):
    def merge_with(self, other): 
         # (This implementation)
```

valohai_yaml/yamlgen/yamlgen.py
```diff
@@ -0,0 +1,179 @@
...
+        image=image,
+        parameters=parameters,
+        inputs=inputs,
+    )
+    config = Config()
+    config.steps[step.name] = step
+    return config
+
+
+def get_current_config(config_path: str) -> Optional[Config]:
```
akx: Is this relevant here?

valohai_yaml/yamlgen/yamlgen.py
```diff
@@ -0,0 +1,179 @@
...
+        return generate_config(
+            relative_source_path=relative_source_path,
+            step=parsed.step,
+            image=DEFAULT_DOCKER_IMAGE,
+            parameters=parsed.parameters,
+            inputs=parsed.inputs,
+        )
+
+
+def get_updated_config(source_path: str, config_path: str) -> (Config, Config):
```
akx: Is this relevant here?

JuhaKiili:
Left the fixes as separate commits.

* Made API surface smaller (moved some of the code to valohai-cli, PR coming)
* Renamed yamlgen to generator
* Moved config merging as class methods

There is a comment in code about this, but I wanted to point out: 

The merging logic for this use-case (parsing .py file for new config) is a bit exotic. I'm slightly troubled that the innocent looking `Config.merge_with()` will not return what people might expect. Personally I can live with it.

valohai_yaml/objs/config.py
```diff
@@ -78,9 +79,9 @@ def get_top_level_parsers(cls):
 
     def serialize(self) -> Any:
         return list(chain(
-            ({'step': step.serialize()} for step in self.steps.values()),
-            ({'endpoint': endpoint.serialize()} for endpoint in self.endpoints.values()),
-            ({'pipeline': pipeline.serialize()} for pipeline in self.pipelines.values()),
+            ({'step': self.steps[key].serialize()} for key in sorted(self.steps)),
+            ({'endpoint': self.endpoints[key].serialize()} for key in sorted(self.endpoints)),
+            ({'pipeline': self.pipelines[key].serialize()} for key in sorted(self.pipelines)),
```
akx: These should probably just sort `.items()`, i.e.
```python
({'step': step.serialize()} for (key, step) in sorted(self.steps.items())),
```
 etc.

valohai_yaml/objs/config.py
```diff
@@ -133,6 +134,15 @@ def get_step_by(self, **kwargs) -> Optional[Step]:
                 return step
         return None
 
+    def merge_with(self, other):
```
akx: What about endpoints and pipelines? This should not just silently leave them unmerged, IMHO.

valohai_yaml/utils/merge.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def _merge_dicts(
+        a: dict,
+        b: dict,
+        merger,
+        copier=lambda v: v,
+        skip_missing_a: bool = False,
+        skip_missing_b: bool = False) -> dict:
+    out = {}
```
akx: This could use the somewhat esoteric incantation `out = type(a)()` to create an object of the same type as the original. That'd allow one to pass in an `OrderedDict` and get one out too.

valohai_yaml/generator/generator.py
```diff
@@ -0,0 +1,122 @@
+import os
```
akx: I'm not quite convinced this YAML generator code should live in `valohai-yaml` at all. 😕 (I'm sorry if I was unclear or incoherent in our balcony discussions about that.)

After all, it will only be able to generate configurations based on an AST that uses that as-of-yet-unreleased module that `valohai-yaml` is only a dependency of.

PR #28, opened 2020-03-20
Config merging (valohai-utils use-case only) + yaml serialization changes

JuhaKiili:


valohai_yaml/objs/merge_mode.py
```diff
@@ -0,0 +1,6 @@
+from enum import Enum
```
akx: This belongs in  `valohai_yaml/enums` or similar; the `objs` package is meant for the objects that make up the YAML files.

valohai_yaml/objs/merge_mode.py
```diff
@@ -0,0 +1,6 @@
...
+
+
+class MergeMode(Enum):
```
akx: Mmmm, these merge modes aren't exactly what I meant. This is still a quasi-dependency to the forthcoming utils package (that is, `valohai-yaml` now has to "know" there is a "PythonParser" thing somewhere that's none of its business to know about).

Looking at the implementations for the current PythonParser merge mode, I understand things aren't necessarily as clear-cut as just "Replace items from other", "Merge items from other" (e.g. the typo case in `Step.merge_...`).

How about one more refactoring of this, replacing `merge_mode` with a `merge_strategy` callable that accepts `self` and `other` and returns the new object? We could then implement a default merge strategy (probably "add/replace new items from other") here. If that sounds sane, I can do the dev legwork.

JuhaKiili: Yeah sounds more legit if the "special merging logic" is moved to `valohai-utils`.

If you are going to do the legwork, please use the already existing `removeyaml2` branch & PR in the `valohai-utils` side.

PR #32, opened 2020-09-17
Allow deployments in pipelines

magdapoppins:
- Add node type "deployment"
- Allow file as edge type

### DIFF
--- a/examples/pipeline-example.yaml
+++ b/examples/pipeline-example.yaml
@@ -43,6 +43,26 @@
       - [batch2.metadata.optimal_learning_rate, train.parameter.learning_rate]
       - [batch2.output.processed-images*, train.input.training-images]
 - pipeline:
+    name: My deployment pipeline
+    nodes:
+      - name: batch1
+        type: execution
+        step: Batch feature extraction
+      - name: train
+        type: execution
+        step: Train model
+      - name: deploy
+        type: deployment
+        deployment: MyDeployment
+        endpoints:
+          - predict-digit
+    edges:
+      #  node.entity_type.entity_name    node.entity_type.entity_name
+      - [batch1.input.training-labels*, train.input.training-labels]
+      - [batch1.metadata.optimal_learning_rate, train.parameter.learning_rate]
+      - [batch1.output.processed-images*, train.input.training-images]
+      - [train.output.model, deploy.file.predict-digit.model]
+- pipeline:
     name: My medium pipeline
     nodes:
       - name: batch1
@@ -76,3 +96,13 @@
       - source: batch2.input.training-labels*
         target: train.input.training-labels
       - [train.output.model.pb, test.input.model]
+
+- endpoint:
+    name: predict-digit
+    description: predict digits from image inputs ("file" parameter)
+    image: tensorflow/tensorflow:1.13.1-py3
+    wsgi: predict_wsgi:predict_wsgi
+    files:
+      - name: model
+        description: Model output file from TensorFlow
+        path: model.pb
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -16,6 +16,17 @@ def test_pipeline(pipeline_config: Config):
         for edge in pipeline_config.pipelines["My little pipeline"].edges
     )
     assert any(
+        (
+            edge.source_node == "train"
+            and edge.source_type == "output"
+            and edge.source_key == "model"
+            and edge.target_node == "deploy"
+            and edge.target_type == "file"
+            and edge.target_key == "predict-digit.model"
+        )
+        for edge in pipeline_config.pipelines["My deployment pipeline"].edges
+    )
+    assert any(
         (edge.source_type == "output" and edge.source_key == "model.pb")
         for edge in pipeline_config.pipelines["My medium pipeline"].edges
     )
--- a/valohai_yaml/objs/pipelines/edge.py
+++ b/valohai_yaml/objs/pipelines/edge.py
@@ -7,7 +7,7 @@ def _split_prop(prop: str) -> List[str]:
     return prop.split('.', 2)
 
 
-edge_types = {'input', 'output', 'parameter', 'metadata'}
+edge_types = {'input', 'output', 'parameter', 'metadata', 'file'}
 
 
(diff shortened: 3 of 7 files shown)


examples/pipeline-example.yaml
```diff
@@ -42,6 +42,33 @@
...
+            endpoints:
+              - name: predict-digit
+                files:
+                  - name: model
+    edges:
+      #  node.entity_type.entity_name    node.entity_type.entity_name
+      - [batch1.input.training-labels*, train.input.training-labels]
+      - [batch1.metadata.optimal_learning_rate, train.parameter.learning_rate]
+      - [batch1.output.processed-images*, train.input.training-images]
+      - [train.output.model, deploy.file.model]
```
akx, line 71: If there are multiple endpoints...
```suggestion
      - [train.output.model, deploy.file.predict-digit.model]
```

examples/pipeline-example.yaml
```diff
@@ -42,6 +42,33 @@
...
+      - name: deploy
+        type: deployment
+        deployment:
+          name: MyDeployment
+          version:
+            name: 2000102.0
+            branch: any
+            commit: none
+            endpoints:
+              - name: predict-digit
```
akx, line 63: This must refer to an existing `endpoint:` in the YAML file. This YAML file doesn't seem to be declaring any endpoints?

examples/pipeline-example.yaml
```diff
@@ -42,6 +42,33 @@
...
+                files:
+                  - name: model
```
akx, line 65: ... and if it refers to an existing endpoint, that endpoint specifies these, not the pipeline step.
```suggestion
```

examples/pipeline-example.yaml
```diff
@@ -42,6 +42,33 @@
...
+      - name: train
+        type: execution
+        step: Train model
+      - name: deploy
+        type: deployment
+        deployment:
+          name: MyDeployment
+          version:
+            name: 2000102.0
+            branch: any
```
akx, line 60: Unnecessary.
```suggestion
```

examples/pipeline-example.yaml
```diff
@@ -42,6 +42,33 @@
...
+            commit: none
```
akx, line 61: Why is this required here for deployment nodes if it's not required for execution nodes? (Execution nodes use the commit the blueprint originates from by default; the same should occur here.)

examples/pipeline-example.yaml
```diff
@@ -42,6 +42,33 @@
...
+        step: Batch feature extraction
+      - name: train
+        type: execution
+        step: Train model
+      - name: deploy
+        type: deployment
+        deployment:
+          name: MyDeployment
+          version:
+            name: 2000102.0
```
akx, line 59: This can't be static, since attempting to create a version with the same name as a pre-existing version would fail, and there's no such concept as updating a deployment version's configuration; they're very much immutable.

If left empty, we could use the same logic we use for suggesting a new version name in the UI (i.e. generate a `YYYYMMDD.X` version).

I do foresee a need for changing that logic to allow for e.g. a prefix – should we allow some placeholders here? Something like `foo-{datestamp}.{counter}`? (@ruksi, ideas?)

```suggestion
```

examples/pipeline-example.yaml
```diff
@@ -42,6 +42,33 @@
...
+      - name: batch1
+        type: execution
+        step: Batch feature extraction
+      - name: train
+        type: execution
+        step: Train model
+      - name: deploy
+        type: deployment
+        deployment:
+          name: MyDeployment
```
akx, line 57: I assume a deployment with this name needs to pre-exist. That's fine.

akx, line 57: Considering all the below requests, could a simpler shape for this be

```suggestion
      - name: deploy
        type: deployment
        deployment: MyDeployment
        endpoints:
          - predict-digit
```
or am I missing something?

akx (changes requested):

examples/pipeline-example.yaml
ruksi: we can make the `name` optional and we always add the `{stamp}.{counter}` either way; as the full name or just as suffix

valohai_yaml/objs/pipelines/edge.py
```diff
@@ -7,7 +7,7 @@ def _split_prop(prop: str) -> List[str]:
     return prop.split('.', 2)
 
 
-edge_types = {'input', 'output', 'parameter', 'metadata'}
+edge_types = {'input', 'output', 'parameter', 'metadata', 'files'}
```
akx, line 10:
```suggestion
edge_types = {'input', 'output', 'parameter', 'metadata', 'file'}
```

tests/test_pipeline.py
```diff
@@ -15,6 +15,17 @@ def test_pipeline(pipeline_config: Config):
         )
         for edge in pipeline_config.pipelines["My little pipeline"].edges
     )
+    assert any(
+        (
+            edge.source_node == "train"
+            and edge.source_type == "output"
+            and edge.source_key == "model"
+            and edge.target_node == "deploy"
+            and edge.target_type == "files"
```
akx, line 24:
```suggestion
            and edge.target_type == "file"
```

examples/pipeline-example.yaml
```diff
@@ -42,6 +42,26 @@
...
+        type: deployment
+        deployment: MyDeployment
+        endpoints:
+          - predict-digit
+    edges:
+      #  node.entity_type.entity_name    node.entity_type.entity_name
+      - [batch1.input.training-labels*, train.input.training-labels]
+      - [batch1.metadata.optimal_learning_rate, train.parameter.learning_rate]
+      - [batch1.output.processed-images*, train.input.training-images]
+      - [train.output.model, deploy.files.predict-digit.model]
```
akx, line 64:
```suggestion
      - [train.output.model, deploy.file.predict-digit.model]
```

akx (changes requested):
As spoken on the other side, also singularize the edge type here? Other than that I think this is a very good minimal solution.

akx (approved):

PR #34, opened 2020-10-23
Add deployment version alias list

magdapoppins:
- Clean up `class Deployment` which I had added while figuring things out while making the deployment in pipeline draft
- Update example YAML
- Update DeploymentNode class
- Emfatten tests to make sure endpoints and aliases come through, just to make sure

Could validation be emstrictened for these array fields?

### DIFF
--- a/examples/pipeline-example.yaml
+++ b/examples/pipeline-example.yaml
@@ -51,17 +51,22 @@
       - name: train
         type: execution
         step: Train model
-      - name: deploy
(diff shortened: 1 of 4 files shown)


akx (changes requested):
Could you make sure this does the right things when the endpoints and aliases lists are missing from the node? They should end up as empty lists in the DeploymentNode object.

akx (approved):
Looks good!

PR #37, opened 2021-03-15
Extend mount objects with `type` and `options`

akx:


ruksi (approved):
adding a couple of values to mount definitions and removing the source validation; both to accommodate the upcoming mount features

lgtm

PR #38, opened 2021-03-15
Replace Travis with Github Actions

akx:


akx:
Once technically approved, I'll flip the required statuses flags in GH settings :)

magdapoppins (approved):
I like it! ✨

PR #42, opened 2021-04-06
Add task-node.yaml schema

akx:
This was missing from the mechanical validator...


ruksi (approved):
adding task node def to schema for validation, lgtm

PR #44, opened 2021-06-14
Drop LegacyParameterMap (a relic from 2018)

akx:
It's pretty safe to assume no one uses the legacy API anymore.


magdapoppins (approved):
LGTM ⭐

PR #45, opened 2021-06-15
Emstricten linting (and some upgrades, etc.)

akx:


ruksi (approved):
well linted

ruksi:
there are some issues with dev deps installation but seems like you are on it already :+1:

akx:
> there are some issues with dev deps installation but seems like you are on it already 👍

Yep, darn out-of-support Python 3.5 making me go through hoops :)

PR #46, opened 2021-06-15
Isort/docstring lints

akx:


tokkoro (approved):
More coherent, looks good.

PR #47, opened 2021-06-15
Typings

akx:
This PR:

* adds py.typed
* adds mypy/flake8-annotations lints
* adds type annotations
* removes a legacy function (hot on the heels of #44) that doesn't even work these days


magdapoppins (approved):
Looks well typed.

PR #48, opened 2021-06-15
Pipeline node actions

akx:


valohai_yaml/objs/pipelines/node_action.py
```diff
@@ -1,8 +1,20 @@
...
 
+from ...lint import LintResult
 from ...utils import listify
 from ..base import Item
 
+WELL_KNOWN_WHENS = {
+    'node-complete',  # Node completed (successfully)
+    'node-starting',  # Node about to start
+    'node-error',  # Node errored
+}
```
ruksi, line 11: it might be worth matching these with our actual node/exec/task statuses

afair, `complete` is any kind of ending in our the `enum` helpers while `success` is only the successful completion? but I could be wrong

also, I think `this node ended in any kind of status` could be worth adding if you can't specify multiple `when`s

akx: Mmm, specifying multiple whens is a good idea, actually. I'll pack that in here too. 

Either way, `complete` and `error` are mapped to our statuses; there is no `success`. `starting` is _ing_ on purpose, because we want the action to be able to pre-empt actually starting the node (by e.g. stopping the whole darned pipeline).

ruksi: yeah, thanks for checking it out

I remember that there was something like `success` or `finished` enum collection status thingy somewhere

ruksi: they are a bit all over the place afair, especially the enum helpers (I recall the statuses themselves are relative consistent)

multiple `when`s might be a good addition

ruksi (approved):
good refactorings and addition of the node actions

lgtm

PR #52, opened 2021-09-28
Do re-exports in __init__ modules

akx:
This will placate mypy without [having to use e.g. `--implicit-reexport`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport) in downstream projects.


ruksi (approved):
more explicit `*` import behavior, lgtm

PR #53, opened 2021-09-30
Import lru_cache from functools

akx:
Fixes compatibility with jsonschema>=4.0.

There's no good reason to have been importing the decorator from jsonschema anyway, since it's been in the stdlib since 3.3.

### DIFF
--- a/valohai_yaml/validation.py
+++ b/valohai_yaml/validation.py
@@ -1,11 +1,11 @@
 import json
 import os
 import re
+from functools import lru_cache
 from typing import IO, List, Union
 
 import yaml
 from jsonschema import Draft4Validator, RefResolver, ValidationError
-from jsonschema.compat import lru_cache
 
 from .excs import ValidationErrors
 from .utils import read_yaml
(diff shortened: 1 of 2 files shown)


valohai_yaml/validation.py
```diff
@@ -1,11 +1,11 @@
 import json
 import os
 import re
+from functools import lru_cache
```
ruksi, line 4: :+1:

ruksi (approved):
lgtm

ruksi:
some test failures

```
valohai_yaml/excs.py:3: in <module>
    import jsonschema
E     File "/opt/hostedtoolcache/Python/3.5.10/x64/lib/python3.5/site-packages/jsonschema/__init__.py", line 56
E       raise AttributeError(f"module {__name__} has no attribute {name}")
E                                                                       ^
E   SyntaxError: invalid syntax
Error: Process completed with exit code 4.
```

akx:
@ruksi

> some test failures

yeah, it seems the version of pip that's being used on py3.5 doesn't understand it can't install jsonschema 4.x (which is properly marked python_requires >= 3.7). Will fix...

PR #54, opened 2021-10-05
Add optional `resources` to endpoint definition

ruksi:
Dropping some groundwork for defining default resource values for Valohai endpoints.

Only consideration was that should we use more generic terms (`min/max`) or Kubernetes terms (`requests/limits`) but I opted for generic terms for data scientists as they shouldn't really be concerned with inner workings of Kubernetes.

In a perfect world, we would use vertical pod autoscaling but that is not an option for all installations, and this definition can extended to configure that scaling.

This can also be extended with other hardware requirements like GPUs and the like.

### DIFF
--- a/valohai_yaml/schema/endpoint.yaml
+++ b/valohai_yaml/schema/endpoint.yaml
@@ -41,6 +41,29 @@ properties:
       Specifies the WSGI application to serve (e.g. a Flask application).
       Specify the module (i.e. `package.app`) or
       the module and the WSGI callable (i.e. `package.app:wsgi_callable`).
+  resources:
+    type: object
+    description: Resource requirements and limits for the endpoint.
+    additionalProperties: false
+    properties:
+      cpu:
+        type: object
+        description: CPU resource requirements and limits in vCPU counts e.g. 0.1 is 10% of one CPU.
+        additionalProperties: false
+        properties:
+          min:
+            type: number
+          max:
+            type: number
+      memory:
+        type: object
+        description: Memory requirements and limits in MB e.g. 100 is 100 MB.
+        additionalProperties: false
+        properties:
+          min:
+            type: number
+          max:
+            type: number
 oneOf:
   - required:
       - server-command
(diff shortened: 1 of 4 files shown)


valohai_yaml/schema/endpoint.yaml
```diff
@@ -41,6 +41,44 @@ properties:
...
+        anyOf:
+          - required:
+              - min
+          - required:
+              - max
+    anyOf:
+      - required:
+          - cpu
+      - required:
+          - memory
```
akx, line 81: What does this specification mean? 🤔 

If it is what I think it is, I don't think we'll need to require either of these to be set if the `resources:` block is present...

ruksi: yeah, my thinking was that if `resources` object is defined, it must have either `cpu` or `memory` defined

but the thing is that if `resources: ` without any properties will evaluate to `None` anyway so it's not strictly required in the current usage

will remove

ruksi: the same can be said with `min` and `max` as they all have `additionalProperties: false`

ruksi: fixed

ruksi: now you can get into a silly state with something like `resources: {}` I guess but that is not a biggie

akx: It's not that silly a state IMO.

ruksi: acceptable, I would say

akx (approved):

valohai_yaml/schema/endpoint.yaml
ruksi: forcing strict structure can help with YAML funnies like

```
- endpoint: 
  resources:
  cpu:
    min: 0.1
    max: 1.0
```

but I think we always have `additionalProperties: false` and `resources: ` evaluates to `None` so it's not an issue

PR #55, opened 2021-10-08
Drop Python 3.5 support

akx:


akx:
> there was some `Build (ubuntu-20.04, 3.5) Expected` lingering in the CI, even after a rebase

Yeah, that was the branch protection configuration. Fixed 🔧

PR #57, opened 2021-11-03
Correct accepted types for Input.keep_directories

akx:
The typing for `Input.keep_directories` implied only booleans are accepted, and that's hardly the case.

In addition, `KeepDirectories.cast()` now accepts `KeepDirectories` values too.

### DIFF
--- a/tests/test_step_parsing.py
+++ b/tests/test_step_parsing.py
@@ -1,4 +1,6 @@
-from valohai_yaml.objs.input import KeepDirectories
+import pytest
+
+from valohai_yaml.objs.input import Input, KeepDirectories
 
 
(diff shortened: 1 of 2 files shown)


magdapoppins (approved):
I'd prefer having more specific values instead of the "coercion" approach here, but since this is already in place, I guess this typing does improve it.

akx:
@magdapoppins 

> I'd prefer having more specific values instead of the "coercion" approach here, but since this is already in place, I guess this typing does improve it.

We need the true/false coercion for old YAML versions.

This just makes it so programmatic users of the library (such as valohai-utils) can "legally" directly pass in a KeepDirectories value.

PR #58, opened 2021-11-03
Correct return type for merge_with

akx:
The return type for merge_with was erroneously left `Item` though we know the merge will always return exactly `T` (whose bound is Item).


ruksi (approved):
lgtm

PR #59, opened 2021-11-03
Do not keep empty Node `actions` in serialized output

akx:
There's no need to have the optional empty list there if the input maybe wasn't there in the first place.

This is currently breaking a test in https://github.com/valohai/valohai-utils/pull/75 since valohai-yaml got bumped to 0.20.0 there.

### DIFF
--- a/valohai_yaml/objs/pipelines/node.py
+++ b/valohai_yaml/objs/pipelines/node.py
@@ -43,6 +43,8 @@ class Node(Item):
     def serialize(self) -> dict:
         ser = super().serialize()
         ser['type'] = self.type
+        if not ser.get('actions'):
+            ser.pop('actions', None)
         return ser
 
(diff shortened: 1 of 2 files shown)


valohai_yaml/objs/pipelines/node.py
```diff
@@ -43,6 +43,8 @@ def parse_qualifying(cls, data: dict) -> 'Node':
     def serialize(self) -> dict:
         ser = super().serialize()
         ser['type'] = self.type
+        if not ser.get('actions'):
+            ser.pop('actions', None)
```
ruksi, line 47: if `actions` falsy or non-existent, **try** to remove `actions` from the dict

the code could signify intent a bit more but that is mainly a personal preference :thinking: 

such a tiny snippet anyway + a test to back it up

ruksi (approved):
lgtm!

PR #60, opened 2021-12-09
Make mypy --strict happy 

akx:
A little strictness never hurt anyone, and I got to use `@overload`!


ruksi (approved):
upgrades, type hint refactoring and python 3.10

lgtm!

PR #61, opened 2021-12-14
Pipeline conversion API

akx:
This makes it easier for downstream users to convert YAML style pipeline definitions to payload style (though YAML style pipelines will be allowed as payloads in the near future anyway).

### DIFF
--- a/valohai_yaml/objs/utils.py
+++ b/valohai_yaml/objs/utils.py
@@ -1,6 +1,7 @@
-import typing
 from collections import OrderedDict
-from typing import Any, Iterable, List, Optional, Type, TYPE_CHECKING, TypeVar
+from typing import Any, Callable, Iterable, List
+from typing import MutableMapping as MutableMappingType
+from typing import Optional, Type, TYPE_CHECKING, TypeVar
 
 from valohai_yaml.types import SerializedDict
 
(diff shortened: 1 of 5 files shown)


valohai_yaml/objs/utils.py
```diff
@@ -15,20 +16,31 @@ def consume_array_of(source: SerializedDict, key: str, type: Type[TItem]) -> Lis
...
+    """
     if source is None:
         return []
     out = []
     for item in source:
         if not isinstance(item, type):
-            raise TypeError(f"{item} not a {type}")
+            if not parse:
+                raise TypeError(f"{item} not a {type}")
+            item = parse(item)
```
ruksi, line 37: should we check that the resulting item is of that `type` or you think that is a overkill?

I mean, at least I assumed that would be the case before I read the implementation

ruksi (changes requested):
one pondering but otherwise lgtm, refactoring YAML conversion to the YAML library

valohai_yaml/objs/utils.py
akx: I was thinking about doing it, and decided it might be somewhat overkill.

Since you were thinking about it too, there's now a type assertion too ✨

ruksi (approved):
is nice

PR #62, opened 2021-12-21
Declarative setup + dev tooling upgrades

akx:
This PR:

* converts setup.py to setup.cfg with https://github.com/akx/setuppy2cfg
* tweaks CI settings some
* upgrades development dependencies for good measure


magdapoppins (approved):
Geronimo!

PR #63, opened 2021-12-22
Correct order of applying overrides to step data when converting nodes

akx:
The overrides need to be applied to the data before it is converted, not afterwards.


magdapoppins (approved):
Looks reasonable!

PR #64, opened 2022-01-20
Sort pipeline nodes and edges when serializing

akx:
This makes comparing serialized content much easier.


akx:
This wasn't a good idea after all. A downstream library can sort things if it needs to.

PR #65, opened 2022-02-07
Add continue-on-error boolean to Pipeline Nodes

tokkoro:
Add optional continue-on-error boolean to each node for https://github.com/valohai/roi/issues/3577

PR #66, opened 2022-02-08
Add on-error enum string to Pipeline Nodes

tokkoro:
I was thorn if this setting should be a `boolean` or `enum`. After more discussion this morning, I see that in near future we might have `on-error: "stop-next"` and `on-error: "retry"`. 

The current behavior when error happens is that all other non finished nodes are stopped, so the default value for `on-error` is "stop-all" and the new option that doesn't care about errors is "continue".

This will replace https://github.com/valohai/valohai-yaml/pull/65


valohai_yaml/objs/pipelines/node.py
```diff
@@ -20,14 +20,19 @@ class Node(Item):
     # `actions` will be set on instance level in subclasses
     actions: List[NodeAction]
 
+    # `on_error` what the pipeline should do when this node is erroneous: "stop-all" default, "continue"
+    on_error: str
```
akx, line 24: I think the in-memory representation for this should be an Enum, like `KeepDirectories` and `MultipleMode` are.

akx (changes requested):
`on_error` is definitely better than the boolean. Could you make the value an Enum, though, instead of an open-ended string?

tokkoro:
Now Roi's api of `GET /api/v0/commits/[HASH]/?include=config` returns a config where the `on_error` keys are `on-error`. So it doesn't change the kebab-case into snake_case. I see the conversion in Item.parse() and tried to add ` data['on_error'] = data.pop('on-error')` into `Node.parse_qualifying`  but I don't know where it should happen. Any pointers where to look @akx ?

akx:
@tokkoro I don't think it _should_ change it into snake case, since the serialization is basically the YAML representation.

tokkoro:
Yes, I was thinking something about some unrelated data transformation down the line that are out of valohai-yaml scope. I fixed the import sorting errors and I think we are good to go in current state.

valohai_yaml/schema/deployment-node.yaml
```diff
@@ -20,6 +20,9 @@ properties:
     type: array
     items:
       "$ref": "./node-action.json"
+  on-error:
+    type: string
+    enum: ["stop-all", "continue"]
```
akx, line 25: Seems `ErrorAction.STOP_NEXT` isn't represented here? (Maybe add a test for it?)
```suggestion
    enum: ["stop-all", "stop-next", "continue"]
```

valohai_yaml/schema/execution-node.yaml
```diff
@@ -8,6 +8,9 @@ properties:
     type: string
   step:
     type: string
+  on-error:
+    type: string
+    enum: ["stop-all", "continue"]
```
akx, line 13: Seems `ErrorAction.STOP_NEXT` isn't represented here? (Maybe add a test for it?)
```suggestion
    enum: ["stop-all", "stop-next", "continue"]
```

valohai_yaml/schema/task-node.yaml
```diff
@@ -8,6 +8,9 @@ properties:
     type: string
   step:
     type: string
+  on-error:
+    type: string
+    enum: ["stop-all", "continue"]
```
akx, line 13: Seems `ErrorAction.STOP_NEXT` isn't represented here? (Maybe add a test for it?)
```suggestion
    enum: ["stop-all", "stop-next", "continue"]
```

akx (changes requested):

akx (approved):

PR #68, opened 2022-02-21
Add step to publish to PyPI if build succeeds

magdapoppins:
For pushes to master where there is a tag, build and push to PyPI.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,21 @@
+name: Publish
+'on':
+  push:
+    tags:
+      - "v[0-9]+"
+    
+jobs:
+  Publish:
+    runs-on: ubuntu-20.04
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+    steps:
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.10'
+      - uses: actions/checkout@v2
+      - run: pip install -U pip setuptools build twine
+      - run: python -m build .
+      - run: twine upload --verbose --repository testpypi dist/*
+        env:
+          TWINE_USERNAME: "__token__"
+          TWINE_PASSWORD: ${{ secrets.TESTPYPI_API_TOKEN }}


.github/workflows/publish.yml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  Publish:
+    runs-on: ubuntu-20.04
+    #if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')# Todo add after testing!
+    steps:
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.10'
+      - uses: actions/checkout@v2
+      - run: pip install -U pip setuptools build
```
akx, line 20:
```suggestion
      - run: pip install -U pip setuptools build twine
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,26 @@
...
+      - run: python -m build .
+      - run: pip install twine
+      - run: |
+          TWINE_USERNAME="__token__" \
+          TWINE_PASSWORD=${{ secrets.TESTPYPI_API_TOKEN }} \
+            exec twine upload --verbose --repository testpypi dist/*
```
akx, line 26:
```suggestion
      - run: twine upload --verbose --repository testpypi dist/*
        env:
          TWINE_USERNAME: __token__"
          TWINE_PASSWORD: ${{ secrets.TESTPYPI_API_TOKEN }}
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,26 @@
+name: Publish
+'on':
+  push:
+    branches:
+      - master
```
akx, line 5: Shouldn't this be removed after testing too?

akx (changes requested):

.github/workflows/publish.yml
magdapoppins: Wonder if there is a on tag option...

akx: https://github.com/omerbenamram/pyevtx-rs/pull/13/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR5-R7

PR #70, opened 2022-02-23
Use real exceptions instead of assertions for certain type checks

akx:
This helps with catching certain exceptions downstream.

### DIFF
--- a/valohai_yaml/objs/config.py
+++ b/valohai_yaml/objs/config.py
@@ -2,6 +2,7 @@ import copy
 from itertools import chain
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple
 
+from valohai_yaml.excs import InvalidType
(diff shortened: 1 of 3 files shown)


valohai_yaml/objs/config.py
```diff
@@ -3,6 +3,7 @@
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple
 
 from valohai_yaml.lint import LintResult
+from ..excs import InvalidType
```
magdapoppins, line 6: Should this be an absolute import instead?

akx: Yes – please see https://github.com/valohai/valohai-yaml/pull/71; let's get that merged and I'll rebase this.

magdapoppins (approved):

PR #72, opened 2022-08-11
Pipeline: fix parameter override by step and node

tokkoro:
Parameter defaults from node should override defaults from step as it's more special use case.


valohai_yaml/pipelines/conversion.py
```diff
@@ -60,8 +60,13 @@ def convert_executionlike_node(self, node: Union[ExecutionNode, TaskNode]) -> Co
         if not step:  # pragma: no cover
             raise ValueError(f"Step {step_name} not found in {self.config}")
         step_data = step.serialize()
-        step_data.update(override)  # TODO: this might need to e.g. merge mappings
```
ruksi, line 63: :+1:

ruksi (approved):
lgtm!

PR #73, opened 2022-08-25
Pipeline parameters

tokkoro:
Add Parameters for Pipelines

### DIFF
--- a/valohai_yaml/objs/pipelines/pipeline.py
+++ b/valohai_yaml/objs/pipelines/pipeline.py
@@ -5,6 +5,7 @@ from valohai_yaml.lint import LintResult
 from valohai_yaml.objs.base import Item
 from valohai_yaml.objs.pipelines.edge import Edge
 from valohai_yaml.objs.pipelines.node import Node
+from valohai_yaml.objs.pipelines.pipeline_parameter import PipelineParameter
 from valohai_yaml.types import LintContext, SerializedDict
 from valohai_yaml.utils.lint import lint_iterables
 
@@ -17,11 +18,13 @@ class Pipeline(Item):
         *,
         name: str,
         nodes: List[Node],
-        edges: List[Edge]
+        edges: List[Edge],
+        parameters: List[PipelineParameter],
     ) -> None:
         self.name = name
         self.nodes = nodes
         self.edges = edges
+        self.parameters = parameters
 
     @property
     def node_map(self) -> OrderedDict:  # type: ignore[type-arg]
@@ -32,11 +35,12 @@ class Pipeline(Item):
         data = data.copy()
         data['edges'] = [Edge.parse(e) for e in data.pop('edges', ())]
         data['nodes'] = [Node.parse_qualifying(n) for n in data.pop('nodes', ())]
+        data['parameters'] = [PipelineParameter.parse(e) for e in data.pop('parameters', ())]
         return super().parse(data)
 
     def lint(self, lint_result: LintResult, context: LintContext) -> None:
         context = dict(context, pipeline=self)
-        lint_iterables(lint_result, context, (self.nodes, self.edges))
+        lint_iterables(lint_result, context, (self.nodes, self.edges, self.parameters))
 
     def get_node_by(self, **kwargs: Any) -> Optional[Node]:
         """Get the first node that matches all the passed named arguments."""
--- a/valohai_yaml/objs/pipelines/pipeline_parameter.py
+++ b/valohai_yaml/objs/pipelines/pipeline_parameter.py
@@ -1,0 +1,67 @@
+from typing import TYPE_CHECKING, List, Optional, Union
+
+from valohai_yaml.lint import LintResult
+from valohai_yaml.objs.base import Item
+from valohai_yaml.objs.utils import check_type_and_listify
+from valohai_yaml.types import LintContext, SerializedDict
+from valohai_yaml.utils.node_socket_utils import split_socket_str
+
+if TYPE_CHECKING:
+    from valohai_yaml.objs import Config, Pipeline
+
+
+class PipelineParameter(Item):
+    """Represents a parameter definition within a pipeline definition."""
+
+    def __init__(
+        self,
+        *,
+        name: str,
+        targets: Union[List[str], str],
+        value: Optional[str] = None,
+        default: Optional[str] = None,
+    ) -> None:
+        self.name = name
+        self.default = default if value is None else value
+        if isinstance(targets, str):
+            self.targets = [targets]
+        else:
+            self.targets = check_type_and_listify(targets, str)
+
+    @classmethod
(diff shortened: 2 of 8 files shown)


valohai_yaml/objs/pipelines/parameter.py
```diff
@@ -0,0 +1,68 @@
...
+
+def get_pipeline_parameter_target_node(target: str) -> str:
+    return target.split(".", 1)[0]
+
+
+def get_pipeline_parameter_target_parameter_name(target: str) -> str:
+    return target.split(".", 1)[1]
+
+
+class Parameter(Item):
```
akx, line 20: I think this should be called PipelineParameter for clarity.

valohai_yaml/objs/pipelines/parameter.py
```diff
@@ -0,0 +1,68 @@
...
+    """Represents a parameter definition within a pipeline definition."""
+
+    def __init__(self,
+                 *,
+                 name: str,
+                 targets: List[str],
+                 value: Optional[str] = None,
+                 default: Optional[str] = None,
+                 ) -> None:
```
akx, line 29: To be in harmony with the rest of the repo:
```suggestion
    def __init__(
        self,
        *,
        name: str,
        targets: List[str],
        value: Optional[str] = None,
        default: Optional[str] = None,
    ) -> None:
```

valohai_yaml/objs/pipelines/pipeline.py
```diff
@@ -17,11 +18,13 @@ def __init__(
         *,
         name: str,
         nodes: List[Node],
-        edges: List[Edge]
+        edges: List[Edge],
+        parameters: List[Parameter]
```
akx, line 22:
```suggestion
        parameters: List[Parameter],
```

valohai_yaml/schema/pipeline-param.yaml
```diff
@@ -0,0 +1,17 @@
...
+  name:
+    type: string
+  targets:
+    oneOf:
+      - type: string
+      - type: array
+        items:
+          type: string
+  target:
+    type: string
```
akx, line 15: This feels a bit complex, but I suppose it's ergonomic...

valohai_yaml/objs/pipelines/parameter.py
```diff
@@ -0,0 +1,68 @@
...
+if TYPE_CHECKING:
+    from valohai_yaml.objs import Config, Pipeline
+
+
+def get_pipeline_parameter_target_node(target: str) -> str:
+    return target.split(".", 1)[0]
+
+
+def get_pipeline_parameter_target_parameter_name(target: str) -> str:
+    return target.split(".", 1)[1]
```
akx, line 17: Hmm... I can't recall having a spec session about this feature (maybe I was on 🏖️ at that point, who knows), but maybe we should support the full edge target spec for pipeline parameters too? I could easily foresee someone wanting to use a pipeline parameter to replace a node input.

akx (changes requested):

valohai_yaml/objs/pipelines/parameter.py
tokkoro: I have been thinking about it, how to make sure the targets can target any properties in nodes and your suggestion resolves it elegantly. I will make it a full edge target.

valohai_yaml/objs/pipelines/parameter.py
tokkoro: I'm reluctant to include namespace in class name, it should be clear from the context. If we change this then I think we should name the other Parameter as ExecutionParameter. Of course this comes from the problem that **parameter** is very generic name for a class but I don't really know if there's a better name for it.

akx: I get where you're coming from with that, but `Parameter` is an entrenched term for execution parameters and task parameters (so renaming it to `ExecutionParameter` wouldn't necessarily be a good idea), and I can foresee it being very easy to accidentally import one `Parameter` when you mean the other `Parameter`...

valohai_yaml/utils/node_socket_utils.py
```diff
@@ -0,0 +1,13 @@
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import List
+
+
+def split_socket_str(socket_str: str) -> 'List[str]':
```
akx, line 7:
```suggestion
from typing import List

def split_socket_str(socket_str: str) -> List[str]:
```

valohai_yaml/utils/node_socket_utils.py
```diff
@@ -0,0 +1,13 @@
...
+    """Split str formatted socket into its parts: node name, socket type, target name."""
```
akx, line 8: Could be a good idea to use the same nomenclature `Edge` uses: node/type/key.

valohai_yaml/utils/node_socket_utils.py
```diff
@@ -0,0 +1,13 @@
...
+    return socket_str.split('.', 2)
+
+
+def get_socket_str(node_name: str, socket_type: str, target_name: str) -> str:
```
akx, line 12: Same here: node/type/key.

valohai_yaml/objs/pipelines/pipeline_parameter.py
```diff
@@ -0,0 +1,66 @@
...
+        self.default = default if value is None else value
+
+    @classmethod
+    def parse(cls, data: SerializedDict) -> 'PipelineParameter':
+        data = data.copy()
+        # targets can be a string or a list of strings
+        if 'target' in data:
+            if 'targets' in data:
+                raise TypeError("Pipeline parameter cannot have both: target and targets")
+            data['targets'] = [data.pop('target', ())]
```
akx, line 35: No need for the default, we definitively know `target` is there.
```suggestion
            data['targets'] = [data.pop('target')]
```

valohai_yaml/objs/pipelines/pipeline_parameter.py
```diff
@@ -0,0 +1,66 @@
...
+        if isinstance(data['targets'], str):
+            data['targets'] = [data['targets']]
```
akx, line 37: I don't think we need to do this cast here since we call `check_type_and_listify` in `__init__`?
```suggestion
```

akx (changes requested):
Minor improvement requests, getting there 😄

valohai_yaml/objs/pipelines/pipeline_parameter.py
tokkoro: I read the listify part of the function name meaning it converts the argument in a list if necessary, like `listify(5) => [5]` but `listify([8, 6]) => [8, 6]`. That wasn't the case and there was no errors as a string was valid iterable input with string type.

PR #76, opened 2022-09-08
Pipeline parameter embetterments

akx:
The serialization change empirically fixes the test failure in https://github.com/valohai/valohai-utils/pull/96 (which expects the pipeline put in to be the same that comes out).

I probably shouldn't have accepted #73 without any tests to begin with, my bad...


tokkoro (approved):
Looks even gooder

PR #79, opened 2022-11-14
Allow using `parameter-value` with falsy but extant values

akx:


ruksi (approved):
when a simple `if` check is enough 🎵

lgtm!

PR #81, opened 2022-12-20
Pipeline: ensure unique node names while linting

tokkoro:
Some inner logic in pipeline handling trusts the names of nodes to be unique so would prevent some hard to track bugs if duplicated nodes are caught early.

### DIFF
--- a/valohai_yaml/objs/pipelines/pipeline.py
+++ b/valohai_yaml/objs/pipelines/pipeline.py
@@ -1,4 +1,4 @@
-from collections import OrderedDict
+from collections import Counter, OrderedDict
 from typing import Any, List, Optional
 
 from valohai_yaml.lint import LintResult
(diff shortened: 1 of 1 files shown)


valohai_yaml/objs/pipelines/pipeline.py
```diff
@@ -45,6 +45,16 @@ def get_data(self) -> SerializedDict:
         return data
 
     def lint(self, lint_result: LintResult, context: LintContext) -> None:
+        # ensure unique names for nodes
+        node_name_counts = {}
+        for node in self.nodes:
+            node_name_counts.setdefault(node.name, 0)
+            node_name_counts[node.name] += 1
```
akx, line 52: Import `itertools.Counter` and
```suggestion
        # ensure unique names for nodes
        node_name_counts = Counter(node.name for node in self.nodes)
```

akx (changes requested):

akx (approved):
🚀

PR #82, opened 2022-12-20
Drop Python 3.6, set up pre-commit, switch to Hatch, etc.

akx:
Yuletide cleanup 🎄 🧹


.pre-commit-config.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - id: isort
+  - repo: https://github.com/PyCQA/flake8
+    rev: 6.0.0
+    hooks:
+      - id: flake8
+        additional_dependencies: &flake8_dependencies
+          - flake8-annotations
+          - flake8-bugbear
+          - flake8-docstrings
+          - flake8-isort
```
ruksi, line 21: is there a reason to have **both** `isort` (introduced through the new pre-hooks) and `flake8-isort` (was previously used as the `isort` wrapper)?

_could_ theoretically cause an issue that `isort` is different version of `flake8-isort` and thus they keep wrangling imports to their desired order to infinity (not a huge risk as `isort` is quite stable)

valohai_yaml/utils/merge.py
```diff
@@ -19,7 +19,7 @@ def merge_dicts(
     out: TD = type(a)()
 
     # Hack to keep the iteration order the same...
-    keys = list(a)  # type: List[Any]
+    keys: List[Any] = list(a)
```
ruksi, line 22: probably belongs to the previous commit but no biggie

ruksi (changes requested):
one question about `isort`

.pre-commit-config.yaml
akx: Oh, yeah, good copypaste catch.

ruksi (approved):
looking stellar 🌃

Issue #80, opened 2022-11-21
Add execution time limits to steps


noorula:
It would be great if one could add the execution time limits to steps with `valohai.yaml`. This is possible in the UI and when starting an execution with the API but when building pipelines it would be be useful not having to go to the UI separately to set this.

akx:
Also done.

PR #83, opened 2022-12-20
Add time-limit and no-output-timeout to step parsing 

akx:
Refs #80 (still needs support elsewhere)

The human-readable duration format is pretty ad-hoc, because the [ISO8601 duration standard](https://en.wikipedia.org/wiki/ISO_8601#Durations) clearly isn't meant for human beings.


valohai_yaml/schema/step.yaml
```diff
@@ -50,3 +50,18 @@ properties:
   environment:
     type: string
     description: Default execution environment (ID or substring of environment name)
+  time-limit:
+    oneOf:
+      - type: integer
+      - type: string
+    description: >
+      The time limit for the step (in seconds or as a string (e.g. "1h 30m 5s")).
```
magdapoppins, line 58: Nit: could drop the () around "e.g. ..." as the double parenthesis look a bit off?

tests/test_step_parsing.py
```diff
@@ -139,3 +141,12 @@ def test_input_extras(input_extras_config):
...
     assert Input(name="foo", keep_directories=value).keep_directories == expected
+
+
+def test_timeouts(timeouts_config):
+    assert timeouts_config.steps['short-time-limit'].time_limit == datetime.timedelta(seconds=300)
+    assert timeouts_config.steps['short-time-limit'].no_output_timeout is None
+    assert timeouts_config.steps['big-no-output-timeout'].time_limit is None
+    assert timeouts_config.steps['big-no-output-timeout'].no_output_timeout.days == 1
+    assert timeouts_config.steps['human-readable-time-limit'].time_limit == datetime.timedelta(seconds=5405)
+    assert timeouts_config.steps['human-readable-time-limit'].no_output_timeout.total_seconds() == 86400 * 2
```
magdapoppins, line 152: Looking at the test fixture and the test, I took a double take at how 
```
- step:
    name: human-readable-time-limit
    image: busybox
    command: foo {parameters}
    time-limit: 1h 30m 5s
    no-output-timeout: 48h
```
leads to 
`assert timeouts_config.steps['human-readable-time-limit'].no_output_timeout.total_seconds() == 86400 * 2`
especially since we mix `total_seconds`, `days` and just the vanilla timedelta in these checks. Is it purposeful to have these different approaches or might it make it more readable 
`...time_limit == datetime.timedelta(seconds=300)`
way in all of these?

valohai_yaml/schema/step.yaml
akx: De-nested! 💪

tests/test_step_parsing.py
akx: Sure. :)

magdapoppins (approved):

PR #84, opened 2023-01-17
Use `ruff` instead of `flake8` and the like

ruksi:
I was setting up a new development environment for some YAML work and though might as well do little something-something 🐶


akx (approved):
Looks good!

PR #85, opened 2023-01-18
Add `node-selector` and `devices` to endpoint definitions

ruksi:
`resources.devices` can be used to specify which device plugin exposed devices the endpoint needs and how many

`node-selector` can be used to further specify device subtype if the nodes are properly labeled i.e. not just some GPU but a specific GPU

`node-selector` can be used alone to just select certain nodes to deploy on; likewise `resources.devices` can also be used to say that the endpoint needs "just some NVIDIA GPU"

### DIFF
--- a/valohai_yaml/schema/endpoint.yaml
+++ b/valohai_yaml/schema/endpoint.yaml
@@ -41,6 +41,10 @@ properties:
       Specifies the WSGI application to serve (e.g. a Flask application).
       Specify the module (i.e. `package.app`) or
       the module and the WSGI callable (i.e. `package.app:wsgi_callable`).
+  node-selector:
+    type: string
+    description: Kubernetes node selector specifies which nodes the deployment should be scheduled to.
+    pattern: "="
   resources:
     type: object
     description: Resource requirements and limits for the endpoint.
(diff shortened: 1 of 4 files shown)


valohai_yaml/schema/endpoint.yaml
```diff
@@ -41,6 +41,10 @@ properties:
       Specifies the WSGI application to serve (e.g. a Flask application).
       Specify the module (i.e. `package.app`) or
       the module and the WSGI callable (i.e. `package.app:wsgi_callable`).
+  node-selector:
+    type: string
+    description: Kubernetes node selector specifies which nodes the deployment should be scheduled to.
+    pattern: "="
```
ruksi, line 47: I considered making it an `object` but opted to go with the string syntax that we use elsewhere (API, UI, admin tooling, etc.)

valohai_yaml/schema/endpoint.yaml
```diff
@@ -64,6 +68,13 @@ properties:
             type: number
           max:
             type: number
+      devices:
+        type: object
+        description: "Devices required e.g. 'nvidia.com/gpu: 1' for one NVIDIA GPU."
+        patternProperties:
+          "^.+$":
+            type: number
+            minimum: 1
```
ruksi, line 77: GPUs would always `integer`s but maybe there are other device types that use floats 🤷

akx (approved):
Looks straightforward enough!

Issue #86, opened 2023-01-23
Allow steps to define icon and category


JuhaKiili:
For the step library project, we want show logos and bling-bling, if the steps are things like 3rd party data platforms.

Example:
```
 - step:
     name: snowflake-query
     image: valohai/dbconnectors
     command: python query.py {parameters}
     icon: https://valohai.com/img/orig/SNOW-35164165.png
     category: Database query
```

PR #87, opened 2023-01-23
Add step icon + category

JuhaKiili:
For the step library project, we want show logos and bling-bling, if the steps are things like 3rd party data platforms.
Screenshot from 2023-01-23 17-20-00

Example:
```
 - step:
     name: snowflake-query
     image: valohai/dbconnectors
     command: python query.py {parameters}
     icon: https://valohai.com/img/orig/SNOW-35164165.png
     category: Database query
```

Fixes https://github.com/valohai/valohai-yaml/issues/86


akx (changes requested):
Looks good but needs a rebase.

akx (approved):

PR #89, opened 2023-01-30
Add support for parameter UI widgets

### LINKED ISSUE #88: Support for parameter UI widgets (YAML)
Support in YAML to ask for a custom UI widget for parameter editing like SQL or DatumAlias.

Example UI:
Image

Example YAML:
```
 - step:
     name: snowflake-query
     image: valohai/dbconnectors
     command: python query.py {parameters}
     icon: https://valohai.com/img/orig/SNOW-35164165.png
     category: Database query
     parameters:
     - name: SQLquery
       type: string
       default: SELECT * FROM table
       widget: sql
     - name: datum_alias
       type: string
       widget: datumalias
```

JuhaKiili:
Example YAML:
```
 - step:
     name: snowflake-query
     image: valohai/dbconnectors
     command: python query.py {parameters}
     icon: https://valohai.com/img/orig/SNOW-35164165.png
     category: Database query
     parameters:
     - name: SQLquery
       type: string
       default: SELECT * FROM table
       widget: sql
     - name: datum_alias
       type: string
       widget: 
         type: datumalias
         settings:
           width: 100
```

Fixes #88

### DIFF
--- a/tests/test_step_parsing.py
+++ b/tests/test_step_parsing.py
@@ -26,7 +26,7 @@ def test_parse(example1_config):
 
     # test that we parsed all params
     parameters = step.parameters
-    assert len(parameters) == 7
+    assert len(parameters) == 9
     # test that we can access them by name
     assert parameters['seed'].default == 1
     assert parameters['decoder-spec'].default == 'gauss'
@@ -39,6 +39,8 @@ def test_parse(example1_config):
         'encoder-layers',
         'denoising-cost-x',
         'decoder-spec',
+        'sql-query',
+        'output-alias',
     ]
 
     # test that `get_step_by` works
@@ -154,3 +156,12 @@ def test_timeouts(timeouts_config):
 def test_bling(example1_config: Config) -> None:
     assert example1_config.steps['run training'].category == "Training"
     assert example1_config.steps['run training'].icon == "https://valohai.com/assets/img/valohai-logo.svg"
+
+
+def test_widget(example1_config: Config) -> None:
+    parameters = example1_config.steps['run training'].parameters
+    assert parameters['sql-query'].widget and parameters['sql-query'].widget.type == "sql"
+    widget = parameters['output-alias'].widget
+    assert widget and widget.type == "datumalias"
+    assert widget and widget.settings and widget.settings["width"] == 123
+    assert parameters['decoder-spec'].widget is None
--- a/valohai_yaml/objs/parameter.py
+++ b/valohai_yaml/objs/parameter.py
@@ -4,6 +4,7 @@ from typing import Any, Iterable, List, Optional, Union
 from valohai_yaml.excs import InvalidType, ValidationErrors
 from valohai_yaml.lint import LintResult
 from valohai_yaml.objs.base import Item
+from valohai_yaml.objs.parameter_widget import ParameterWidget
 from valohai_yaml.types import LintContext, SerializedDict
 from valohai_yaml.utils import listify
 
@@ -48,7 +49,8 @@ class Parameter(Item):
         pass_false_as: Optional[str] = None,
         choices: Optional[Iterable[ValueAtomType]] = None,
         multiple: Optional[Union[str, MultipleMode]] = None,
-        multiple_separator: str = ','
+        multiple_separator: str = ',',
+        widget: Optional[ParameterWidget] = None,
     ) -> None:
         self.name = name
         self.type = type
@@ -62,6 +64,8 @@ class Parameter(Item):
         self.choices = (list(choices) if choices else None)
         self.multiple = MultipleMode.cast(multiple)
         self.multiple_separator = str(multiple_separator or ',')
+        self.widget = widget
+
         self.default = (listify(default) if self.multiple else default)
         if self.type == 'flag':
             self.optional = True
@@ -193,6 +197,13 @@ class Parameter(Item):
             return _format_atom(value)
 
         raise NotImplementedError(f'unknown multiple type {self.multiple!r}')
+
+    @classmethod
+    def parse(cls, data: SerializedDict) -> 'Parameter':
+        kwargs = data.copy()
+        if 'widget' in data:
+            kwargs['widget'] = ParameterWidget.parse(data.get('widget'))
+        return super().parse(kwargs)
 
     def lint(
         self,
--- a/valohai_yaml/objs/parameter.py
+++ b/valohai_yaml/objs/parameter.py
@@ -202,7 +202,7 @@ class Parameter(Item):
     def parse(cls, data: SerializedDict) -> 'Parameter':
         kwargs = data.copy()
         if 'widget' in data:
-            kwargs['widget'] = ParameterWidget.parse(data.get('widget'))
+            kwargs['widget'] = ParameterWidget.parse(data['widget'])
         return super().parse(kwargs)
 
     def lint(
--- a/valohai_yaml/objs/parameter_widget.py
+++ b/valohai_yaml/objs/parameter_widget.py
@@ -1,0 +1,38 @@
+from typing import Optional, Union
+
+from valohai_yaml.objs.base import Item
+from valohai_yaml.types import SerializedDict
+
+
+class ParameterWidget(Item):
+    """An UI widget for editing parameter values."""
+
+    def __init__(
+        self,
+        *,
+        type: str,
+        settings: Optional[dict] = None,
+    ) -> None:
+        self.type = str(type).lower()
+        self.settings = dict(settings or {})
+
+    def serialize(self) -> Optional[Union[SerializedDict, str]]:
+        if self.settings:
+            return {
+                'type': self.type,
+                'settings': self.settings,
+            }
+        return self.type
+
+    @classmethod
+    def parse(cls, data: Optional[Union['ParameterWidget', dict, str]]) -> 'ParameterWidget':
+        if isinstance(data, dict):
+            return cls(
+                type=data['type'],
+                settings=data.get('settings'),
+            )
+
+        if isinstance(data, cls):
+            return data
+
+        return cls(type=str(data))
--- a/valohai_yaml/objs/parameter_widget.py
+++ b/valohai_yaml/objs/parameter_widget.py
@@ -25,7 +25,7 @@ class ParameterWidget(Item):
         return self.type
 
     @classmethod
-    def parse(cls, data: Optional[Union['ParameterWidget', dict, str]]) -> 'ParameterWidget':
+    def parse(cls, data: Union['ParameterWidget', dict, str]) -> 'ParameterWidget':
         if isinstance(data, dict):
             return cls(
                 type=data['type'],
--- a/valohai_yaml/schema/param-item.yaml
+++ b/valohai_yaml/schema/param-item.yaml
@@ -78,3 +78,16 @@ properties:
     description: |
       The separator to use when interpolating multiple values into a command parameter.
       Defaults to a comma.
+  widget:
+    oneOf:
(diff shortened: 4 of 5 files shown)


valohai_yaml/schema/param-item.yaml
```diff
@@ -78,3 +78,12 @@ properties:
     description: |
       The separator to use when interpolating multiple values into a command parameter.
       Defaults to a comma.
+  widget:
+    type: string
+    default: none
+    enum:
+      - none
+      - sql
+      - datumalias
```
akx, line 87: Let's not lock the given values down here? It's always a pain in the rear to have to update the whole ecosystem depending on this library.
```suggestion
```

Also, I think we should look into the future a bit and allow this to be an object as well as a string:

```yaml
widget: sql
```
would be equivalent to
```yaml
widget:
  type: sql
```
and in the future each widget might have other settings too?

Parsing would always result in a dict or None, serializing would attempt the shortest form, i.e. if the dict in the Step object has `len(widget) == 1 and widget.get("type")`, then it'd be serialized back to a string.

akx (changes requested):

valohai_yaml/objs/parameter_widget.py
```diff
@@ -0,0 +1,28 @@
...
+    def cast(cls, value: Optional[Union['ParameterWidget', OrderedDict, dict, str]]) -> Optional['ParameterWidget']:
+        if not value:
+            return None
+        if isinstance(value, (dict, OrderedDict)):
+            return cls(type=value['type'])
+        if isinstance(value, cls):
+            return value
+        value = str(value).lower()
+        if value == 'none':
+            return None
```
akx, line 27: I wonder if the special `"none"` value is really necessary at all..?

tests/test_step_parsing.py
```diff
@@ -154,3 +156,10 @@ def test_timeouts(timeouts_config):
 def test_bling(example1_config: Config) -> None:
     assert example1_config.steps['run training'].category == "Training"
     assert example1_config.steps['run training'].icon == "https://valohai.com/assets/img/valohai-logo.svg"
+
+
+def test_widget(example1_config: Config) -> None:
+    assert example1_config.steps['run training'].parameters['sql-query'].widget and \
+        example1_config.steps['run training'].parameters['sql-query'].widget.type == "sql"
+    assert example1_config.steps['run training'].parameters['output-alias'].widget and \
+        example1_config.steps['run training'].parameters['output-alias'].widget.type == "datumalias"
```
akx, line 165: DRY... :)

```suggestion
    parameters = example1_config.steps['run training'].parameters
    assert parameters['sql-query'].widget and parameters['sql-query'].widget.type == "sql"
    assert parameters['output-alias'].widget and parameters['output-alias'].widget.type == "datumalias"
```

akx, line 165: Can you also check that `widget` is `None` as expected for some other parameter?

valohai_yaml/objs/parameter_widget.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __init__(
+        self,
+        *,
+        type: str,
+    ) -> None:
+        self.type = type.lower()
+
+    @classmethod
+    def cast(cls, value: Optional[Union['ParameterWidget', OrderedDict, dict, str]]) -> Optional['ParameterWidget']:
```
akx, line 18: Also, ah, `cast` is only a thing for our `Enum`s; this should be `parse()` (and you should also define `serialize()`, so ParameterWidgets with just a type can be serialized back to a simple `str`.

akx (changes requested):

valohai_yaml/objs/parameter_widget.py
```diff
@@ -0,0 +1,39 @@
...
+    """An UI widget for editing parameter values."""
+
+    def __init__(
+        self,
+        *,
+        type: str,
+        settings: Optional[dict] = None,
+    ) -> None:
+        self.type = type.lower()
+        self.settings = settings
```
akx, line 17: Might be nicer that `self.settings` always was a thing, even if empty.

```suggestion
        self.settings = dict(settings or {})
```

This doesn't need changes to the serializer though since empty dicts are falsy ✨

valohai_yaml/objs/parameter_widget.py
```diff
@@ -0,0 +1,39 @@
...
+                'type': self.type,
+                'settings': self.settings,
+            }
+        return self.type
+
+    @classmethod
+    def parse(cls, data: Union['ParameterWidget', dict, str]) -> 'ParameterWidget':
+        if isinstance(data, dict):
+            return cls(
+                type=data.get('type', ''),
```
akx, line 31: I think `type` is required?
```suggestion
                type=data['type'],
```

valohai_yaml/objs/parameter_widget.py
```diff
@@ -0,0 +1,39 @@
...
+                settings=data.get('settings'),
+            )
+
+        if isinstance(data, cls):
+            return data
+
+        value = str(data).lower()
```
akx, line 38: What's the reason for the lower-casing? And why isn't it done for the other style of `type`?

valohai_yaml/schema/param-item.yaml
```diff
@@ -1,3 +1,4 @@
+# Here is the full schema
```
akx, line 1: I'm not sure this comment has much value..?

valohai_yaml/schema/param-item.yaml
```diff
@@ -78,3 +79,17 @@ properties:
...
       Defaults to a comma.
+  widget:
+    oneOf:
+      - type: object
+        required:
+        - type
+        properties:
+          type:
+            type: string
+            default: none
```
akx, line 90: I don't think there should be a default..?
```suggestion
```

akx (changes requested):

valohai_yaml/schema/param-item.yaml
JuhaKiili: Leftover from my dialog with ChatGPT ;)

valohai_yaml/objs/parameter_widget.py
JuhaKiili: Not doing it for the dict was my error. Pushed a fix now.

The reasoning for lowercase was that people will use `SQL`, `Sql` and `sql` and I don't see the upside of making this case sensitive.

akx: Fair enough.

valohai_yaml/objs/parameter_widget.py
```diff
@@ -0,0 +1,39 @@
...
+            return cls(
+                type=str(data['type']).lower(),
+                settings=data.get('settings'),
+            )
+
+        if isinstance(data, cls):
+            return data
+
+        value = str(data).lower()
+        return cls(type=value)
```
akx, line 39: Since you're `.lower`ing this in the constructor anyway,
```suggestion
        return cls(type=data)
```

valohai_yaml/objs/parameter_widget.py
```diff
@@ -0,0 +1,39 @@
...
+                'type': self.type,
+                'settings': self.settings,
+            }
+        return self.type
+
+    @classmethod
+    def parse(cls, data: Union['ParameterWidget', dict, str]) -> 'ParameterWidget':
+        if isinstance(data, dict):
+            return cls(
+                type=str(data['type']).lower(),
```
akx, line 31: You're `.lower`ing the type in the constructor anyway, so
```suggestion
                type=data['type'],
```

valohai_yaml/objs/parameter_widget.py
```diff
@@ -0,0 +1,39 @@
...
+class ParameterWidget(Item):
+    """An UI widget for editing parameter values."""
+
+    def __init__(
+        self,
+        *,
+        type: str,
+        settings: Optional[dict] = None,
+    ) -> None:
+        self.type = type.lower()
```
akx, line 16:
```suggestion
        self.type = str(type).lower()
```

valohai_yaml/objs/parameter.py
```diff
@@ -199,11 +203,16 @@ def lint(
         lint_result: LintResult,
         context: LintContext,
     ) -> None:
+        prefix = f'Step {context["step"].name}, parameter {self.name}'
+        self.lint_flags(prefix, lint_result)
+        self.lint_default(prefix, lint_result)
```
akx, line 208: What's the reason for this change/refactoring, by the way? I don't see it described in the commit message or PR description...

akx (changes requested):

valohai_yaml/objs/parameter.py
```diff
@@ -62,6 +64,8 @@ def __init__(
         self.choices = (list(choices) if choices else None)
         self.multiple = MultipleMode.cast(multiple)
         self.multiple_separator = str(multiple_separator or ',')
+        self.widget = ParameterWidget.parse(widget)
```
akx, line 67: Just
```suggestion
        self.widget = widget
```
here, and call `ParameterWidget.parse` in `Parameter`'s `parse`, á la

```python
    @classmethod
    def parse(cls, data: SerializedDict) -> 'Parameter':
        kwargs = data.copy()
        kwargs['widget'] = ParameterWidget.parse(data.get('widget'))
        return cls(**kwargs)
```

akx (changes requested):

valohai_yaml/objs/parameter.py
JuhaKiili: Sorry! I was rushing it and had a lukihäiriö. 

It should be kosher now.

valohai_yaml/objs/parameter_widget.py
```diff
@@ -0,0 +1,38 @@
...
+    def serialize(self) -> Optional[Union[SerializedDict, str]]:
+        if self.settings:
+            return {
+                'type': self.type,
+                'settings': self.settings,
+            }
+        return self.type
+
+    @classmethod
+    def parse(cls, data: Optional[Union['ParameterWidget', dict, str]]) -> 'ParameterWidget':
```
akx, line 28: Based on the code (I swear it had `if data is None` in an earlier revision?!) `Optional` is no longer true here; this would do the wrong thing for `None`.

Either get rid of that –

```suggestion
    def parse(cls, data: Union['ParameterWidget', dict, str]) -> 'ParameterWidget':
```

or alternately make it return `None` for falsy values...

```suggestion
    def parse(cls, data: Optional[Union['ParameterWidget', dict, str]]) -> Optional['ParameterWidget']:
        if not data:
            return None
```

I'd go with option 2, because it would simplify `Parameter.parse` too; you wouldn't need to check whether `widget` is in `data`...

akx (changes requested):
Sorry – one more tweak...

valohai_yaml/objs/parameter_widget.py
JuhaKiili: I was going for option 2 earlier, but the mypy error made me think its not cool (?)

```
Return type "Optional[ParameterWidget]" of "parse" incompatible with return type "ParameterWidget" in supertype "Item"  [override]
```

I've pushed option 1 as separate commit now.

akx: Alright! I'd forgotten we had strictly specified `parse` needs to return the thing.

akx (approved):
LGTM! Thanks for putting up with the viilaus :)

JuhaKiili:
> LGTM! Thanks for putting up with the viilaus :)

No problem. It was all valid. I was a bit lost with how the serializing worked.

PR #94, opened 2023-05-15
Improve error messages for indentation errors

### LINKED ISSUE #49: Indentation errors are cryptically linted
`vh lint` throws an error if the valohai.yaml file has lines that are not properly indented.

The error message is cryptic:
`error:   Type validation on pipeline: None is not of type 'object' (2.pipeline)`

The error message should clearly say that the issue is in indentation.

Below an example of a pipeline that fails. But if you indent everything under pipeline (lineno 2->) it's valid.

```yaml
- pipeline:
  name: mypipeline
  nodes:
    - name: validation-node
      type: task
      step: validate
    - name: ingest-node
      type: task
      step: ingest
      override:
      inputs:
          - name: image
  edges:
    - [validation-node.parameter.image, ingest-node.parameter.image]
```

memona008:
Resolves #49

## Why does this pull request exist?
`vh lint` was not giving proper errors in case of indentation errors

## Changes made in this pull request:

- Added pyyaml exception handling to get indentation errors with their location details
- Added human readable hints about indentation correction if user gives valid but unindented yaml

### DIFF
--- a/valohai_yaml/lint.py
+++ b/valohai_yaml/lint.py
@@ -22,6 +22,9 @@ class LintResult:
     def add_warning(self, message: str, location: None = None, exception: Optional[Exception] = None) -> None:
         self.messages.append({'type': 'warning', 'message': message, 'location': location, 'exception': exception})
 
+    def add_hint(self, message: str, location: None = None, exception: Optional[Exception] = None) -> None:
+        self.messages.append({'type': 'hint', 'message': message, 'location': location, 'exception': exception})
+
     @property
     def warning_count(self) -> int:
         return sum(1 for m in self.messages if m['type'] == 'warning')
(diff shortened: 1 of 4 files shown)


valohai_yaml/lint.py
```diff
@@ -60,7 +64,14 @@ def lint_file(file_path: str) -> LintResult:
...
 
-    data = read_yaml(yaml)
+    try:
+        data = read_yaml(yaml)
+    except pyyaml.YAMLError as err:
+        if hasattr(err, 'problem_mark'):
+            mark = err.problem_mark
+            indent_error = f"Indentation Error at line {mark.line + 1}, column {mark.column + 1}"
+            raise ValidationError(indent_error) from err
+
```
akx, line 74: I think we should reraise if we didn't hit the `problem_mark` branch :)
```suggestion
        raise

```

valohai_yaml/lint.py
```diff
@@ -79,6 +90,11 @@ def lint(yaml: YamlReadable) -> LintResult:
         styled_message = style(error.message, fg='red')
         styled_path = style('.'.join(obj_path), bold=True)
         lr.add_error(f'  {styled_validator} validation on {styled_schema_path}: {styled_message} ({styled_path})')
+        # when path has only 2 nodes. it means it has problem in main steps/pipelines/endpoints objects
+        if len(error.path) == 2 and not error.instance:
+            styled_hint = style(f'File contains valid yaml but there might be an indentation '
+                                f'error in following configuration: {styled_path}', fg='blue')
+            lr.add_hint(styled_hint)
```
akx, line 97:
```suggestion
            styled_hint = style(
                "File contains valid YAML but there might be an indentation "
                f"error in following configuration: {styled_path}",
                fg='blue',
            )
            lr.add_hint(styled_hint)
```

Easier to read when less indented, and YAML is an acronym :D

akx (changes requested):
Couple inline comments, and could we also get a test that tries these out?

akx (approved):

PR #96, opened 2023-05-25
Add mode to Execution node to merge/override inputs

euvolen-valohai:
- update schema and obj of Execution node
- add private methods to PipelineConvector to handle merging
- update tests and conftests

ref [#2377](https://github.com/valohai/roi/issues/2377)


tests/conftest.py
```diff
@@ -9,6 +9,7 @@
 endpoint_config = config_fixture('endpoint-example.yaml')
 optional_default_param_config = config_fixture('optional-default-param.yaml')
 pipeline_config = config_fixture('pipeline-example.yaml')
+pipeline_overriden_config = config_fixture('pipeline-overriden-example.yaml')
```
akx, line 12: Typo: [`overridden`](https://dictionary.cambridge.org/dictionary/english/overridden)

valohai_yaml/objs/pipelines/execution_node.py
```diff
@@ -1,30 +1,51 @@
...
 from typing import Any, Dict, List, Optional, Union
 
 from valohai_yaml.lint import LintResult
 from valohai_yaml.objs.pipelines.node import ErrorAction, Node
 from valohai_yaml.objs.pipelines.node_action import NodeAction
-from valohai_yaml.types import LintContext, NodeOverrideDict
+from valohai_yaml.types import LintContext, NodeOverrideDict, SerializedDict
+
+
+class Mode(Enum):
```
akx, line 10: This is a too generic name – `OverrideMode`?

I'd probably also put it in a module of its own (`valohai_yaml.objs.pipelines.override_mode`?)

valohai_yaml/objs/pipelines/execution_node.py
```diff
@@ -1,30 +1,51 @@
...
+    OVERRIDE = 'override'  # default: override inputs in a node
+
+    @classmethod
+    def cast(cls, value: Optional[Union['Mode', str]]) -> 'Mode':
+        if not value:
+            return Mode.OVERRIDE
+        if isinstance(value, Mode):
+            return value
+        value = str(value).lower()
+        return Mode(value)
```
akx, line 23: A note for later: we might want to refactor the now practically quadruplicate enum cast code to one.

valohai_yaml/objs/pipelines/execution_node.py
```diff
@@ -1,30 +1,51 @@
...
-        on_error: Union[str, ErrorAction] = ErrorAction.STOP_ALL,
+            self,
+            *,
+            name: str,
+            step: str,
+            actions: Optional[List[NodeAction]] = None,
+            override: Optional[NodeOverrideDict] = None,
+            on_error: Union[str, ErrorAction] = ErrorAction.STOP_ALL,
+
+            mode: Union[str, Mode] = Mode.OVERRIDE
```
akx, line 41: Please fix your IDE to not do this reindentation :)

Also, add a trailing comma?

valohai_yaml/schema/execution-node.yaml
```diff
@@ -15,6 +15,9 @@ properties:
     type: array
     items:
       "$ref": "./node-action.json"
+  mode:
```
akx, line 18: This should also be named better, e.g. `override-mode`?

valohai_yaml/pipelines/conversion.py
```diff
@@ -60,8 +93,15 @@ def convert_executionlike_node(self, node: Union[ExecutionNode, TaskNode]) -> Co
         if not step:  # pragma: no cover
             raise ValueError(f"Step {step_name} not found in {self.config}")
         step_data = step.serialize()
-        step_data.update(override)
 
+        if node_data.get("mode"):
```
akx, line 97: Just look at the node directly :)
```suggestion
        if node.override_mode == OverrideMode.MERGE:
```
or similar.

valohai_yaml/pipelines/conversion.py
```diff
@@ -60,8 +93,15 @@ def convert_executionlike_node(self, node: Union[ExecutionNode, TaskNode]) -> Co
...
+            inputs = override.pop("inputs", {})
+            inputs = [OrderedDict(input) for input in inputs] if inputs else []
+            merged_inputs = self._merge_overridden_inputs(inputs, step_data.get("inputs", []))
+            #todo: this is a hack to get the extend inputs but it needs to be extended to other fields as well
```
akx, line 101: So... this only works for `inputs` right now? (I mean, changing `parameters` defaults is also a very valid use case.) That should be probably made clearer somehow.

In fact, maybe for future proofing, we should already make `override_mode` a mapping á la

```yaml
override-mode:
  inputs: override
  parameters: merge
```
– WDYT? Too much complexity?

valohai_yaml/pipelines/conversion.py
```diff
@@ -52,6 +53,38 @@ def convert_deployment_node(self, node: DeploymentNode) -> ConvertedObject:
...
+        for ordered_dict1 in overridden:
+            name = ordered_dict1['name']
+            if name not in dict1_map:
+                merged_list.append(ordered_dict1)
+
+        return merged_list
+
+    def _merge_defaults(self, default1: Optional[Any], default2: Optional[Any]) -> List[str]:
+        default1 = [default1] if isinstance(default1, str) else (default1 if default1 else [])
+        default2 = [default2] if isinstance(default2, str) else (default2 if default2 else [])
```
akx, line 85: Would `valohai_yaml.utils.listify` help you here?

akx, line 85: I think these two `merge` functions could be made free functions and maybe moved to `pipelines/overrides.py` or something?

valohai_yaml/pipelines/conversion.py
```diff
@@ -52,6 +53,38 @@ def convert_deployment_node(self, node: DeploymentNode) -> ConvertedObject:
         node_data["aliases"] = node.aliases
         return node_data
 
+    def _merge_overridden_inputs(self, overridden: List[OrderedDict], list2: List[OrderedDict]) -> List[OrderedDict]:
```
akx, line 56: `list2` is not the best name for a variable I've ever seen 😁

akx (changes requested):
There's a bunch of notes/thoughts within, might be worth to read through them before fixing one, because I think we could end up with a much more elegant solution here.

Since this is touching the YAML schema too, it's worth to spend some additional time here to make it right from the get-go :)

valohai_yaml/objs/pipelines/execution_node.py
euvolen-valohai: Agree about name, OverrideMode sounds much better. When it comes to location, we have ErrorActions inside node.py so I did same, but I am not sure. Maybe all localish field should be in file  eg util_classes or smth like it?

valohai_yaml/pipelines/conversion.py
euvolen-valohai: can do, but is it really a use case? it should works for all variables there. I think it is too complex.

euvolen-valohai:
Closed to be replaced with another [approach](https://github.com/valohai/valohai-yaml/pull/97)

PR #97, opened 2023-05-29
Merge step inputs with overridden node inputs

euvolen-valohai:
- create overridden-properties.yaml schema to keep override props
- add merge_inputs method
- add test

Simplify  #96


examples/pipeline-override-example.yaml
```diff
@@ -0,0 +1,45 @@
...
+        override:
+             inputs:
+                - name: training-labels
+                  default: "merged node label"
+                  optional: true
+             parameters:
+                - name: a
+                  default: "merged node parameter A"
+                - name: d
+                  default: "merged node parameter D"
```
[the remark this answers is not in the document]
euvolen-valohai: it doesn't right now. I will add validation into lint function.

valohai_yaml/objs/pipelines/execution_node.py
```diff
@@ -42,3 +43,13 @@ def get_parameter_defaults(self) -> Dict[str, Any]:
             in self.override['parameters']
             if 'default' in param
         }
+
+    def get_input_defaults(self) -> Dict[str, Any]:
```
akx, line 47: Wouldn't this have to be
```suggestion
    def get_input_defaults(self) -> Dict[str, List[str]]:
```

valohai_yaml/objs/step.py
```diff
@@ -105,6 +106,14 @@ def get_parameter_defaults(self, include_flags: bool = True) -> Dict[str, Any]:
             if parameter.default is not None and (include_flags or parameter.type != 'flag')
         }
 
+    def get_input_defaults(self) -> Dict[str, Any]:
```
akx, line 109: Same here..?

```suggestion
    def get_input_defaults(self) -> Dict[str, List[str]]:
```

valohai_yaml/objs/step.py
```diff
@@ -105,6 +106,14 @@ def get_parameter_defaults(self, include_flags: bool = True) -> Dict[str, Any]:
...
+        """Get a dict mapping parameter names to their defaults (if set)."""
```
akx, line 110: Copy-pasted docstring? :)

valohai_yaml/pipelines/conversion.py
```diff
@@ -55,21 +54,25 @@ def convert_deployment_node(self, node: DeploymentNode) -> ConvertedObject:
...
         node_data = node.serialize()
         step_name = node_data.pop("step")
-        override = node_data.pop("override", {})
         step = self.config.get_step_by(name=step_name)
         if not step:  # pragma: no cover
             raise ValueError(f"Step {step_name} not found in {self.config}")
+
         step_data = step.serialize()
-        step_data.update(override)
+        node_data.pop("override", {})
```
akx, line 62: Won't this now drop everything else that could have been overridden? (`image`, `command`, ..?)

valohai_yaml/lint.py
```diff
@@ -65,6 +65,53 @@ def lint_file(file_path: str) -> LintResult:
...
+        if pipeline_data:
+            pipeline_name = pipeline_data.get('name')
+            if pipeline_name:
+                pipeline_nodes = pipeline_data.get('nodes', [])
+                pipeline_dict[pipeline_name] = {
+                    i['name']: {i['step']: i.get('override', {})}
+                    for i in pipeline_nodes if 'step' in i
+                }
+
+    return pipeline_dict, step_dict
```
akx, line 112: Ah, I think this validation (if it does what I think it does) should maybe be done in `Pipeline`'s `lint()`, where everything has already been parsed. You can access the entirety of the config via `context["config"]` there.

Let me know if I've misunderstood what's going on here...

akx (dismissed):
Some thoughts to begin with 😅

If nothing else, I think we can do that new validation much easier..?

valohai_yaml/pipelines/conversion.py
```diff
@@ -55,21 +54,25 @@ def convert_deployment_node(self, node: DeploymentNode) -> ConvertedObject:
...
+
+        inputs_from_node = node.get_input_defaults()
+        inputs_from_step = step.get_input_defaults()
+        step_data["inputs"] = inputs_from_step
+        step_data["inputs"].update(inputs_from_node)
```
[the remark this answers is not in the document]
euvolen-valohai: I tried to make logics as it was in parameters but I like this enhancing. I am rewriting it.

valohai_yaml/pipelines/conversion.py
euvolen-valohai: Yeah, my bad. Will fix asap

valohai_yaml/objs/step.py
euvolen-valohai: 🤦🏻‍♂️ I won't be ever copypasting again...

valohai_yaml/pipelines/conversion.py
```diff
@@ -55,22 +54,25 @@ def convert_deployment_node(self, node: DeploymentNode) -> ConvertedObject:
...
+
+        inputs_from_node = node.get_input_defaults()
+        inputs_from_step = step.get_input_defaults()
+        step_data["inputs"] = {**inputs_from_step, **inputs_from_node}
 
         parameters_from_node = node.get_parameter_defaults()
         parameters_from_step = step.get_parameter_defaults(include_flags=True)
-        step_data["parameters"] = parameters_from_step
-        step_data["parameters"].update(parameters_from_node)
+        step_data["parameters"] = {**parameters_from_step, **parameters_from_node}
```
akx, line 69: Hmm... It looks like `get_*_defaults()` doesn't return inputs or parameters for which no defaults have been set? Additionally, the value returned from the dict is just a list of defaults, not the input or parameter definition (as I suppose we expect..?)

I may be missing something here though. 😅

akx (changes requested):
I rebased this for ease of review and slightly cleaned up the lint function (less nesting and so on), but I now have my doubts about the usage of `get_*_defaults()`, see comment within 🤔

valohai_yaml/pipelines/conversion.py
euvolen-valohai: I will check it

euvolen-valohai: well, I though that idea is to override values so if value in override but not in step, it added it into node. So imho it should keep working that way, but I also can miss smth here. @tokkoro do I?
I updated tests for covering this use case.

valohai_yaml/objs/pipelines/execution_node.py
```diff
@@ -36,16 +36,16 @@ def lint(self, lint_result: LintResult, context: LintContext) -> None:
...
+        for input_name in self.get_serialized_inputs():
             if input_name not in step_inputs:
                 lint_result.add_error(f"{error_prefix}: input {input_name} does not exist in step")
-        for parameter_name in self.get_parameter_defaults():
+        for parameter_name in self.get_serialized_parameters():
             if parameter_name not in step_parameters:
                 lint_result.add_error(f"{error_prefix}: parameter {parameter_name} does not exist in step")
 
-    def get_parameter_defaults(self) -> Dict[str, Any]:
+    def get_serialized_parameters(self) -> Dict[str, Any]:
```
akx, line 48: I... don't see why this was renamed? Especially since this does return a parameter-to-default mapping, and not serialized `Parameter` objects, and even so it only returns those that have `"default"` set?

valohai_yaml/objs/pipelines/execution_node.py
```diff
@@ -55,7 +55,7 @@ def get_parameter_defaults(self) -> Dict[str, Any]:
             if 'default' in param
         }
 
-    def get_input_defaults(self) -> Dict[str, List[str]]:
+    def get_serialized_inputs(self) -> Dict[str, List[str]]:
```
akx: Ditto – this rename doesn't seem correct.

valohai_yaml/objs/step.py
```diff
@@ -106,13 +106,23 @@ def get_parameter_defaults(self, include_flags: bool = True) -> Dict[str, Any]:
             if parameter.default is not None and (include_flags or parameter.type != 'flag')
         }
 
-    def get_input_defaults(self) -> Dict[str, List[str]]:
+    def get_serialized_parameters(self, include_flags: bool = True) -> Dict[str, Any]:
```
akx, line 109: Same here – this seems like a bad name for the function.

valohai_yaml/objs/step.py
```diff
@@ -106,13 +106,23 @@ def get_parameter_defaults(self, include_flags: bool = True) -> Dict[str, Any]:
...
+        """Get a dict mapping parameter names to their defaults (if set)."""
```
akx, line 110: This seems copy-pasted from the other implementation, and as such isn't really correct (since this implementation will return a `''` for non-set defaults.

valohai_yaml/objs/step.py
```diff
@@ -106,13 +106,23 @@ def get_parameter_defaults(self, include_flags: bool = True) -> Dict[str, Any]:
...
+        return {
+            name: parameter.default if parameter.default is not None and (include_flags or parameter.type != 'flag')
+            else ''
+            for (name, parameter)
+            in self.parameters.items()
+
+        }
+
+    def get_serialized_inputs(self) -> Dict[str, List[str]]:
```
akx, line 119: And same here...

akx (changes requested):
I don't think the last commit makes sense, unfortunately – and it's also renaming a public API (`get_parameter_defaults` is used by other pieces of software).

If you wanted a dict with all parameters with e.g. `None` substituting for no default value, then you could do

```python
{**dict.fromkeys(step.parameters, None), **step.get_parameter_defaults(include_flags=True)}
```
to merge the nones dict with defaults when available.

However, I'm still wondering whether the final value will be correct, since `override`s are, as we know, supposed to be merged with a step definition?

valohai_yaml/pipelines/conversion.py
tokkoro: Isn't this the place where we generate the template for nodes? the template should include only name-value pairs in dictionary for parameters and not the definition. Roi will find them in the commit that is defined in node and then add these values on those when creating the actual Execution.

tokkoro: So to summarize, you two are talking about two different things here. @akx I think you are thinking of a different place, as this is the template for execution/task node and shouldn't include parameter definitions (those are found by commit+step_name of the node). @euvolen-valohai you are correct about the override logic, that if something has default just in node override and not in the step it should have the overriding default value here after.

valohai_yaml/objs/pipelines/execution_node.py
```diff
@@ -3,42 +3,55 @@
...
+        for input_name in self.override.inputs:
+            if input_name not in step_inputs:
+                lint_result.add_error(f"{error_prefix}: input {input_name} does not exist in step")
+        for parameter_name in self.override.parameters:
+            if parameter_name not in step_parameters:
+                lint_result.add_error(f"{error_prefix}: parameter {parameter_name} does not exist in step")
 
-    def get_parameter_defaults(self) -> Dict[str, Any]:
-        if "parameters" not in self.override:
+    def get_parameters_defaults(self) -> Dict[str, Any]:
```
akx, line 49: Please don't rename this public API function.

valohai_yaml/objs/pipelines/execution_node.py
```diff
@@ -3,42 +3,55 @@
...
-
+    override: Override
     def __init__(
         self,
         *,
         name: str,
         step: str,
         actions: Optional[List[NodeAction]] = None,
-        override: Optional[NodeOverrideDict] = None,
+        override: Optional[SerializedDict] = None,
```
akx, line 21: Why the type change? If anything, this should allow an `Override` object too now that we have one.

```suggestion
        override: Optional[Union[Override, NodeOverrideDict]] = None,
```

valohai_yaml/objs/pipelines/execution_node.py
```diff
@@ -3,42 +3,55 @@
...
+from valohai_yaml.objs.pipelines.override import Override
+from valohai_yaml.types import LintContext, SerializedDict
 
 
 class ExecutionNode(Node):
     """Represents an execution node within a pipeline definition."""
 
     type = 'execution'
-
+    override: Override
```
akx, line 14: `override` is probably the less common case, so it might be better to just not have one if it would be empty?
```suggestion
    override: Optional[Override]
```
(Also requires the relevant change in the constructor to make `.override` None if an empty dict is passed in.)

valohai_yaml/objs/pipelines/override.py
```diff
@@ -0,0 +1,57 @@
...
+from valohai_yaml.objs.base import Item
+from valohai_yaml.objs.environment_variable import EnvironmentVariable
+from valohai_yaml.objs.input import Input
+from valohai_yaml.objs.utils import check_type_and_dictify, check_type_and_listify, serialize_into
+from valohai_yaml.types import SerializedDict
+from valohai_yaml.utils.parse import parse_common_kwargs
+
+
+class Override(Item):
+    """Represents a Override class within an Execution Node."""
```
akx, line 14:
```suggestion
    """Represents the step fields overridden in an execution or task node."""
```
might be more accurate?

valohai_yaml/utils/parse.py
```diff
@@ -0,0 +1,17 @@
...
+
+from valohai_yaml.objs.environment_variable import EnvironmentVariable
+from valohai_yaml.objs.input import Input
+from valohai_yaml.objs.mount import Mount
+from valohai_yaml.objs.parameter import Parameter
+from valohai_yaml.objs.utils import consume_array_of
+from valohai_yaml.types import SerializedDict
+
+
+def parse_common_kwargs(data: SerializedDict) -> Dict[str, Any]:
```
akx, line 11: Mmm, these are common to... steps and overrides, I guess? Could we find a better name and place for this?

valohai_yaml/objs/pipelines/override.py
```diff
@@ -0,0 +1,57 @@
...
+    def __init__(
+        self,
+        *,
+        image: Optional[str] = '',
+        command: Optional[Union[List[str], str]] = '',
+        parameters: Optional[Iterable[Parameter]] = (),
+        inputs: Optional[Iterable[Input]] = (),
+        mounts: Optional[Iterable[Mount]] = (),
+        environment_variables: Optional[Iterable[EnvironmentVariable]] = (),
+        environment: Optional[str] = None,
```
akx, line 25: The default values here don't look exactly right. I think all of them should default to `None`, as in "not set in the `override:` stanza".

valohai_yaml/objs/pipelines/override.py
```diff
@@ -0,0 +1,57 @@
...
+    @classmethod
+    def parse(cls, data: SerializedDict) -> 'Override':
+        kwargs = parse_common_kwargs(data)
+        inst = cls(**kwargs)
+        inst._original_data = data
+        return inst
+
+
+
+    def serialize(self) -> OrderedDict:  # type: ignore[type-arg]
```
akx, line 44: Similarly, this should take care to only serialize the values that are set (not `None`).

valohai_yaml/objs/step.py
```diff
@@ -105,6 +101,7 @@ def get_parameter_defaults(self, include_flags: bool = True) -> Dict[str, Any]:
             if parameter.default is not None and (include_flags or parameter.type != 'flag')
         }
 
+
```
akx, line 104: Spurious added line?

valohai_yaml/pipelines/conversion.py
```diff
@@ -55,22 +55,24 @@ def convert_deployment_node(self, node: DeploymentNode) -> ConvertedObject:
...
         node_data = node.serialize()
         step_name = node_data.pop("step")
-        override = node_data.pop("override", {})
+
         step = self.config.get_step_by(name=step_name)
         if not step:  # pragma: no cover
             raise ValueError(f"Step {step_name} not found in {self.config}")
         step_data = step.serialize()
-        step_data.update(override)
+        override = node_data.pop("override", None)
```
akx, line 63: Should this use `Override.parse()`..?

valohai_yaml/pipelines/conversion.py
```diff
@@ -55,22 +55,24 @@ def convert_deployment_node(self, node: DeploymentNode) -> ConvertedObject:
...
 
-        parameters_from_node = node.get_parameter_defaults()
-        parameters_from_step = step.get_parameter_defaults(include_flags=True)
-        step_data["parameters"] = parameters_from_step
-        step_data["parameters"].update(parameters_from_node)
+        all_inputs = override.pop("inputs", []) + step_data.pop("inputs", [])
+        merged_inputs = [next(key for key in all_inputs if key["name"] == name) for name in step.inputs]
+        step_data['inputs'] = {input['name']: listify(input.get('default', None)) for input in merged_inputs}
```
akx, line 67: I'm not following what's happening in this code at all 😅

valohai_yaml/pipelines/conversion.py
```diff
@@ -55,22 +55,24 @@ def convert_deployment_node(self, node: DeploymentNode) -> ConvertedObject:
...
 
-        step_data["inputs"] = {
-            i["name"]: listify(i.get("default")) for i in step_data.get("inputs", [])
-        }
+        all_parameters = override.pop("parameters", []) + (step_data.pop("parameters", []))
+        merged_parameters = [next(key for key in all_parameters if key["name"] == name) for name in step.parameters]
+        step_data['parameters'] = {parameter['name']: parameter.get('default', '') for parameter in merged_parameters}
```
akx, line 71: Again, I really don't think this is correct, since a Step doesn't have a

```yaml
parameters:
  foo: 123
  bar: quux
```
construct.

akx (changes requested):

valohai_yaml/utils/parse.py
euvolen-valohai: parse_step_properties(or _fields)? It is 'step' because override works with it. What do you think?

valohai_yaml/pipelines/conversion.py
euvolen-valohai: it would be wrong format, cos step_data is serialised. node_data has a serialised override so inputs and parameters can be easily merged

valohai_yaml/pipelines/conversion.py
euvolen-valohai: Code takes a list of possible names of inputs/parameters from step, then simply takes first item of all_(array) with the same name. I should probably rename vars for the good

valohai_yaml/objs/pipelines/execution_node.py
euvolen-valohai: if it is optional, it would lean to overcomplexity in lint()

valohai_yaml/objs/pipelines/execution_node.py
euvolen-valohai: Override.parse as a classmethod works with type Dict[str, Any] so Union no need imho

akx: The point is you should be able to pass in an Override you've built by hand if you're using this as a programmatic API (like e.g. https://github.com/valohai/papi does).

valohai_yaml/objs/pipelines/execution_node.py
akx: You can refactor `lint()` to call out to a separate `_lint_override()` if an override is present.

valohai_yaml/utils/parse.py
akx: `parse_common_step_properties` sounds pretty good. Please also add a docstring then! :)

valohai_yaml/pipelines/conversion.py
akx: Yes, but you could parse the Override, merge things using the parsed objects (as opposed to dicts) and then serialize back?

valohai_yaml/objs/pipelines/execution_node.py
tokkoro: @euvolen-valohai seems that this is implemented if so resolve the conversation :+1: same for other renamings.

akx:
I went ahead and rebased this on top of the now-merged split #103 (the original branch is `eugene/feature/override-input-simple-backup-2023-08-03`).

valohai_yaml/objs/pipelines/override.py
```diff
@@ -69,3 +72,38 @@ def serialize(self) -> OrderedDict:  # type: ignore[type-arg]
                     elide_empty_iterables=True,
                 )
         return val
+
+    @classmethod
+    def default_merge(cls, a: Override | None, b: Any) -> Override:
```
akx, line 77: The name `default_merge` has a special meaning; it's supposed to be a merger for two objects of the same type.

If you don't mean to interfere with the `merge_with` mechanism here, maybe use another name (`apply_defaults_from_step` if that's what this is supposed to do?) and correct the names and types of the arguments too (and the locals – `result`?)

valohai_yaml/objs/pipelines/override.py
```diff
@@ -69,3 +72,38 @@ def serialize(self) -> OrderedDict:  # type: ignore[type-arg]
...
+        )
+        result.environment_variables = merge_dicts(
+            b.environment_variables,
+            result.environment_variables,
+            merger=merge_simple,
+            copier=copy.deepcopy,
+        )
+        return result
+
+    def normalize_to_template(self) -> OrderedDict:
```
akx, line 99: This would require a docstring at the very least. It could also just modify a `template` dict passed in?

valohai_yaml/objs/pipelines/override.py
```diff
@@ -69,3 +72,38 @@ def serialize(self) -> OrderedDict:  # type: ignore[type-arg]
...
+        override = self.serialize()
+        override["inputs"] = {
+            input["name"]: listify(input.get("default", None))
+            for input in override.get("inputs", [])
+        }
+        override["parameters"] = {
+            parameter["name"]: parameter.get("default", None)
+            for parameter in override.get("parameters", [])
+        }
```
akx, line 108: Again, this would probably be neater if it used the objects instead of just yoloing around with the serialized dicts :)

Something like

```suggestion
        inputs = {
            input.name: listify(input.default)
            for input in self.inputs
        }
        parameters = {
            parameter.name: parameter.default
            for parameter in override.parameters
            if parameter.default is not None
        }
```
or similar...?

akx (changes requested):
As mentioned, I split out the Override object stuff to #103, and that was reviewed and merged by @ruksi – this became that much simpler to review, but I'm not still 1000% sure of things here.

PR #98, opened 2023-05-30
Add task template

euvolen-valohai:
Solve [issue](https://github.com/valohai/roi/issues/2895)
- add schemas
- add classes Task and VariantParameter, enums TaskParameterType and VariantParameterStyle
- update configs
- add tests


examples/step-with-tasks.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+- task:
+    step: run training
+    name: task 1 run training
+    type: bayesian_tpe
+    parameters:
+      - name: A
+        style: single
+        rules:
+          value: 25
```
tokkoro, line 23: This is the technical format we use, but I'm not sure if it's necessary to have rules object separated and could be flatten to parameter level? Everyone saves typing.
```suggestion
    parameters:
      - name: A
        style: single
        value: 25
```

examples/step-with-tasks.yaml
```diff
@@ -0,0 +1,57 @@
...
+    inputs:
+      - name: training-set-images
+        default: http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz
+        description: the image data for the training set
+    parameters:
+      - name: B
+        type: integer
+        default: 1
+
+- task:
```
tokkoro, line 15: I think this example should include every (relevant) type of variant parameter we have:

> SingleVariant
> MultipleVariant
> LinearVariant
> LogspaceVariant
> RandomVariant

(These are names of the classes not values for style)

valohai_yaml/objs/task.py
```diff
@@ -0,0 +1,94 @@
...
+    def cast(cls, value: Optional[Union['VariantParameterStyle', str]]) -> 'VariantParameterStyle':
+        if not value:
+            return VariantParameterStyle.SINGLE
+        if isinstance(value, VariantParameterStyle):
+            return value
+        value = str(value).lower()
+        return VariantParameterStyle(value)
+
+
+class VariantParameter(Item):
```
tokkoro, line 49: What do you think, at some point we want to move these from Roi to Valohai-yaml fully. As for example we now use Parameters for Executions from valohai-yaml library. Should that be a part of this PR or should we do it later?

valohai_yaml/objs/task.py
```diff
@@ -0,0 +1,94 @@
...
+    check_type_and_listify,
+    consume_array_of,
+)
+
+
+class TaskParameterType(Enum):
+    """Represents a task parameter style definition."""
+
+    BAYESIAN_TPE = 'bayesian_tpe'
+    GREED_SEARCH = 'greed_search'
```
tokkoro, line 15: is this supposed to be Grid Search? I don't think we have Greed search or have I missed something?

valohai_yaml/schema/task.yaml
```diff
@@ -0,0 +1,28 @@
+type: object
+required:
+  - name
+  - step
+  - parameters
+additionalProperties: false
+properties:
```
tokkoro, line 7: Should we have (optional) task configuration being part of this or have you planned it to come later?

For every Task we have `stop_expression` text and `maximum_queued_executions` integer
For random search we also have `execution_count` integer
and for Bayesian we have 
- `execution_count` integer
- `execution_batch_size` integer
- `optimization_target_metric` text
- `optimization_target_value` float
- `engine` text

tokkoro (changes requested):
Looking this from users perspective and also had couple of scoping questions is there a plan to include them later or should we have them here.

examples/step-with-tasks.yaml
euvolen-valohai: imho, it will be much harder to validate => more chance to make a mistake. I planned to write schemas for all rules, but    it i s too complex I am open to suggestions :)

valohai_yaml/objs/task.py
euvolen-valohai: frankly I don't think it is a good idea to do more then one things in one pull request. I would do it in another PR

valohai_yaml/objs/task.py
euvolen-valohai: Why did the greedy internet user become a search engine expert?
Because they wanted to hoard all the "greed-able" information!

valohai_yaml/schema/task.yaml
euvolen-valohai: Should not this all be in step template? I thought to make a task template more related on step and to add only task-specific property into template.

examples/step-with-tasks.yaml
tokkoro: Do you have plans to match `rules` per `style`? So that it only accepts specific `rules` format when `style` is "single"? I think that should be as easy or even easier when you flat the structure with [oneOf](https://json-schema.org/understanding-json-schema/reference/combining.html#oneof) and having every rule in separated collection.

valohai_yaml/objs/task.py
tokkoro: Sounds good, just make sure to read through the current implementation in Roi to make sure there isn't any blockers in this version.

valohai_yaml/schema/task.yaml
tokkoro: These should be task specific configurations nothing to do with the step.

valohai_yaml/objs/config.py
```diff
@@ -110,6 +115,7 @@ def lint(  # type: ignore[override]
             self.steps,
             self.endpoints,
             self.pipelines,
+            self.tasks
```
akx, line 118:
```suggestion
            self.tasks,
```

valohai_yaml/objs/task.py
```diff
@@ -0,0 +1,109 @@
...
+
+from valohai_yaml.objs.base import Item
+from valohai_yaml.objs.utils import (
+    check_type_and_listify,
+    consume_array_of,
+)
+
+
+class TaskParameterType(Enum):
+    """Represents a task parameter style definition."""
```
akx, line 12: Misnamed - this doesn't have much to do with parameters. 

```suggestion
class TaskType(Enum):
    """Represents a task type."""
```

valohai_yaml/objs/task.py
```diff
@@ -0,0 +1,109 @@
...
+    MANUAL_SEARCH = 'manual_search'
+
+    @classmethod
+    def cast(cls, value: Optional[Union['TaskParameterType', str]]) -> 'TaskParameterType':
+        if not value:
+            return TaskParameterType.GRID_SEARCH
+        if isinstance(value, TaskParameterType):
+            return value
+        value = str(value).lower()
+        return TaskParameterType(value)
```
akx, line 27:
```suggestion
    def cast(cls, value: TaskType | str | None) -> TaskType:
        if not value:
            return TaskType.GRID_SEARCH
        if isinstance(value, TaskType):
            return value
        value = str(value).lower()
        return TaskType(value)
```

valohai_yaml/objs/task.py
```diff
@@ -0,0 +1,109 @@
...
+        if isinstance(value, VariantParameterStyle):
+            return value
+        value = str(value).lower()
+        return VariantParameterStyle(value)
+
+
+class VariantParameter(Item):
+    """Represents a variant parameter definition."""
+
+    rules: Any
```
akx, line 52: I think we'll know `rules` will always be a dict with string keys, so
```suggestion
    rules: dict[str, Any]
```

valohai_yaml/objs/task.py
```diff
@@ -0,0 +1,109 @@
+from enum import Enum
```
akx, line 1: Adding the annotations future will let you use more modern type syntax in this brand new file. (At some point we'll probably add that to the rest of the codebase.)

You can run something like `ruff --select=UP006,UP007 --python-version=py311 valohai_yaml/objs/task.py` then, to have ruff autofix the old-style `Optional`s etc. for you.

```suggestion
from __future__import annotations

from enum import Enum
```

valohai_yaml/objs/task.py
```diff
@@ -0,0 +1,109 @@
...
+        name: str,
+        style: Optional[Union['VariantParameterStyle', str]] = None,
+    ) -> None:
+        self.rules = rules
+        self.name = name
+        self.style = VariantParameterStyle.cast(style)
+
+
+class Task(Item):
+    """Represents an task definition."""
```
akx, line 69:
```suggestion
    """Represents a task definition."""
```

valohai_yaml/schema/task.yaml
```diff
@@ -0,0 +1,40 @@
...
+    description: The step to run with.
+  optimization_target_metric:
+    type: string
+    description: The step to run with.
+  optimization_target_value:
+    type: float
+    description: The step to run with.
+  engine:
+    type: string
+    description: The step to run with.
```
akx, line 26: Copy-pasted descriptions abound :)

valohai_yaml/schema/variant-param.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - style
+additionalProperties: false
+properties:
+  name:
+    type: string
+  style:
+    type: string
+    enum: [ "logspace", "multiple", "linear", "single", "random" ]
+  rules:
+    $ref: "./rule-item.json"
```
akx, line 15: Could this file have a more specific name (`variant-param-rule-item.json` maybe?)

valohai_yaml/objs/task.py
```diff
@@ -0,0 +1,109 @@
...
+    execution_batch_size: Optional[int]
+    optimization_target_metric: Optional[str]
+    optimization_target_value: Optional[float]
+    engine: Optional[str]
+    def __init__(
+        self,
+        *,
+        name: str,
+        step: str,
+        type: Optional[Union['TaskParameterType', str]] = None,
```
akx, line 85:
```suggestion
        type: TaskType | str | None = None,
```

valohai_yaml/objs/task.py
```diff
@@ -0,0 +1,109 @@
...
+        parameters: Optional[List[VariantParameter]] = None,
+        execution_count: Optional[int] = None,
+        execution_batch_size: Optional[int] = None,
+        optimization_target_metric: Optional[str] = None,
+        optimization_target_value: Optional[float] = None,
+        engine: Optional[str] = None,
+    ) -> None:
+        self.name = name
+        self.step = step
+        self.type = TaskParameterType.cast(type)
```
akx, line 95:
```suggestion
        self.type = TaskType.cast(type)
```

valohai_yaml/objs/task.py
```diff
@@ -0,0 +1,109 @@
...
+    """Represents a variant parameter definition."""
+
+    rules: Any
+    name: str
+    style: VariantParameterStyle
+
+    def __init__(
+        self,
+        *,
+        rules: Optional[Any] = None,
```
akx, line 59:
```suggestion
        rules: dict[str, Any] | None = None,
```

akx (changes requested):
Some initial comments! Couple of typos, etc. but the idea looks good to me 😄

valohai_yaml/schema/task.yaml
euvolen-valohai: 😰

valohai_yaml/objs/task.py
euvolen-valohai: it was copilot this time :)

akx (approved):
I rebased this to get the Prettier stuff from #99 here, and addressed the few remaining comments.

Looking good 🚀

PR #102, opened 2023-07-07
Update JSON Schema to Draft 2020-12

akx:
Removes the hack added by #101 (since the top-level `id` is now `$id`)


axelauvinen (approved):
👍

PR #103, opened 2023-08-03
Make Override a first-class citizen, not just a dict

akx:
Extracted from #97.


valohai_yaml/objs/pipelines/override.py
```diff
@@ -0,0 +1,71 @@
...
+        self.parameters = check_type_and_dictify(parameters, Parameter, "name")
+        self.environment_variables = check_type_and_dictify(
+            environment_variables,
+            EnvironmentVariable,
+            "name",
+        )
+
+    @classmethod
+    def parse(cls, data: SerializedDict) -> Override:
+        kwargs = parse_common_step_properties(data)
```
ruksi, line 47: reuse from the `step.parse`, good good yes yes

ruksi (approved):
give pipeline node step overrides proper types in JSON Schema and Python

lgtm :100:

PR #107, opened 2023-08-14
Pipeline Parameter Conversion

memona008:
Resolves [Pipeline parameters issue](https://github.com/valohai/roi/issues/4500)

## Why does this pull request exist? 
While creating pipeline from valohai-cli (adhoc or not), pipeline parameters were not being included in payload.

## Changed made in this pull request
- Add parameters in payload
- Add pipeline parameters conversion function which converts parameters dict into config-expression structure

### DIFF
--- a/tests/test_pipeline_conversion.py
+++ b/tests/test_pipeline_conversion.py
@@ -50,3 +50,24 @@ def test_pipeline_conversion_override_inputs(pipeline_overridden_config: Config)
     assert overridden["template"]["inputs"].get('training-images', [])[0] == 'overridden node image'
     assert len(overridden["template"]["inputs"].get('training-images', [])) == 1
     assert len(overridden["template"]["parameters"].items()) == 3
+
+
+def test_pipeline_parameter_conversion(pipeline_with_parameters_config):
+    parameter_name = "id"
+    for _name, pipe in pipeline_with_parameters_config.pipelines.items():
+        result = PipelineConverter(
+            config=pipeline_with_parameters_config,
+            commit_identifier="latest",
+        ).convert_pipeline(pipe)
+        assert isinstance(result["parameters"], dict)
+        assert result["parameters"][parameter_name]
+
+        parameter = result["parameters"][parameter_name]
+        assert parameter["config"]["targets"]
+        assert "target" not in parameter["config"]
+        assert isinstance(parameter["config"]["targets"], list)
+
+        # When pipeline parameter has no default value, the expression should be empty
+        parameter_config = next(param for param in pipe.parameters if param.name == parameter_name)
+        expression_value = parameter_config.default if parameter_config.default else ""
+        assert parameter["expression"] == expression_value
--- a/valohai_yaml/pipelines/conversion.py
+++ b/valohai_yaml/pipelines/conversion.py
@@ -1,4 +1,4 @@
-
+import sys
 from typing import Any, Dict, List, Union
 
 from valohai_yaml.objs import (
@@ -7,11 +7,24 @@ from valohai_yaml.objs import (
     ExecutionNode,
     Node,
     Pipeline,
+    PipelineParameter,
     TaskNode,
 )
 from valohai_yaml.objs.pipelines.override import Override
 
 ConvertedObject = Dict[str, Any]
+
+
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+    class ConvertedPipeline(TypedDict):
+        """TypedDict for converted Pipeline object."""
+
+        edges: List[ConvertedObject]
+        nodes: List[ConvertedObject]
+        parameters: Dict[str, ConvertedObject]
+else:
+    ConvertedPipeline = ConvertedObject
 
 
 class PipelineConverter:
@@ -26,10 +39,21 @@ class PipelineConverter:
         self.config = config
         self.commit_identifier = commit_identifier
 
-    def convert_pipeline(self, pipeline: Pipeline) -> Dict[str, List[ConvertedObject]]:
+    def convert_pipeline(self, pipeline: Pipeline) -> ConvertedPipeline:
         return {
             "edges": [edge.get_expanded() for edge in pipeline.edges],
             "nodes": [self.convert_node(node) for node in pipeline.nodes],
+            "parameters": {
+                parameter.name: self.convert_parameter(parameter)
+                for parameter in pipeline.parameters
+            },
+        }
+
+    def convert_parameter(self, parameter: PipelineParameter) -> ConvertedObject:
+        """Convert a pipeline parameter to a config-expression payload."""
+        return {
+            "config": {**parameter.serialize()},
+            "expression": parameter.default if parameter.default else "",
         }
 
     def convert_node(self, node: Node) -> ConvertedObject:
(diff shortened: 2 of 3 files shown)


memona008:
Looking into the solution of above typing errors.

valohai_yaml/pipelines/conversion.py
```diff
@@ -7,29 +7,53 @@
...
     TaskNode,
 )
 from valohai_yaml.objs.pipelines.override import Override
 
 ConvertedObject = Dict[str, Any]
 
 
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+    class ConvertedPipeline(TypedDict):
```
tokkoro, line 20: No change required from this commend.

`ConvertedPipeline` sounds like a pipeline that it self has been converted and not that the data presenting the same pipeline is converted into another form. I see that this follows the `ConvertedObject` naming convention but I would like to discuss if better name could be found, like `PipelineDict` or `PipelineData` or `SerializedPipeline`.

I had a short poll at the office and the end thought was that the `ConvertedPipeline` is the best option.

valohai_yaml/pipelines/conversion.py
```diff
@@ -7,29 +7,53 @@
...
 
     def __init__(
-        self,
-        *,
-        config: Config,
-        commit_identifier: str,
+            self,
+            *,
+            config: Config,
+            commit_identifier: str,
```
tokkoro, line 37: I don't think change is necessary to get this feature out.

Would be nice to have a ruff rule to check the indention of function parameters.

valohai_yaml/pipelines/conversion.py
```diff
@@ -53,9 +77,13 @@ def convert_deployment_node(self, node: DeploymentNode) -> ConvertedObject:
...
 
-    def convert_executionlike_node(self, node: Union[ExecutionNode, TaskNode]) -> ConvertedObject:
+    def convert_executionlike_node(
+            self, node: Union[ExecutionNode, TaskNode]
+    ) -> ConvertedObject:
         node_data = node.serialize()
-        node_data.pop('override', {})  # we'll use the actual object, not the serialization
+        node_data.pop(
+            "override", {}
+        )  # we'll use the actual object, not the serialization
```
tokkoro, line 86: Is there an actual change hidden somewhere? I don't notice. Would be good to changes to format as a separated commit.

valohai_yaml/pipelines/conversion.py
```diff
@@ -64,7 +92,9 @@ def convert_executionlike_node(self, node: Union[ExecutionNode, TaskNode]) -> Co
 
         runtime_config = step_data.setdefault("runtime_config", {})
         if "no-output-timeout" in step_data:
-            runtime_config.setdefault("no_output_timeout", step_data.pop("no-output-timeout"))
+            runtime_config.setdefault(
+                "no_output_timeout", step_data.pop("no-output-timeout")
+            )
```
tokkoro, line 97: Lets leave this out of this PR as it's a no-change.

tests/test_pipeline_conversion.py
```diff
@@ -50,3 +50,18 @@ def test_pipeline_conversion_override_inputs(pipeline_overridden_config: Config)
...
+
+
+def test_pipeline_parameter_conversion(pipeline_with_parameters_config):
+    for _name, pipe in pipeline_with_parameters_config.pipelines.items():
+        result = PipelineConverter(
+            config=pipeline_with_parameters_config,
+            commit_identifier="latest",
+        ).convert_pipeline(pipe)
+        assert isinstance(result["parameters"], dict)
+        parameter_name = "id"
```
tokkoro, line 62: This assign here inside the loop makes me wonder why we need to set it every iteration so what is changing it and I can't find any. As it seems to be constant it could be initialized once out side of the loop.

tests/test_pipeline_conversion.py
```diff
@@ -50,3 +50,18 @@ def test_pipeline_conversion_override_inputs(pipeline_overridden_config: Config)
...
+        assert result["parameters"][parameter_name]
+
+        parameter = result["parameters"][parameter_name]
+        assert parameter["config"]["targets"] or parameter["config"]["target"]
```
tokkoro, line 66: At this point the target should have a name of `targets` and shouldn't have `target` in it at all. PipelineParameter parse method should take care of it.

We should also check that the targets is a list.

```suggestion
        assert parameter["config"]["targets"]
        assert not parameter["config"]["target"]
        assert isinstance(parameter["config"]["targets"], list)
```

Luckily, we have all these cases covered in `pipeline_with_parameters_config`

tokkoro (changes requested):
Small things I noted, would be nice if you can remove style changes from this PR and make the small fix to the test.

valohai_yaml/pipelines/conversion.py
memona008: these changes came from black command. I ran manually on my local.

tests/test_pipeline_conversion.py
memona008: makes sense

tests/test_pipeline_conversion.py
tokkoro: Would be cool to have the test that target is removed:
        `assert not parameter["config"]["target"]`
I don't see this in the latest version.

valohai_yaml/pipelines/conversion.py
```diff
@@ -26,10 +39,21 @@ def __init__(
...
+                parameter.name: self.convert_parameter(parameter)
+                for parameter in pipeline.parameters
+            },
+        }
+
+    def convert_parameter(self, parameter: PipelineParameter) -> ConvertedObject:
+        """Convert a pipeline parameter to a config-expression payload."""
+        return {
+            "config": {**parameter.serialize()},
+            "expression": parameter.default,
```
tokkoro, line 56: Lets mimic the way the web UI works and put empty string here when the default is missing.

tests/test_pipeline_conversion.py
```diff
@@ -50,3 +50,26 @@ def test_pipeline_conversion_override_inputs(pipeline_overridden_config: Config)
...
+        assert parameter["config"]["targets"]
+        assert "target" not in parameter["config"]
+        assert isinstance(parameter["config"]["targets"], list)
+
+        # parameter expression check
+        if _name == "Example Pipeline with Parameters":
+            assert parameter["expression"] == 123
+        else:
+            # When a parameter's default is missing, it should be an empty string
+            assert parameter["expression"] == ""
```
tokkoro, line 75: This works as well, but it speaks for a different cause-effect relation. The name of pipeline doesn't effect the default value straight. You should dig the actual parameter config from pipeline's config and use that to determinate what value the expression should have:

```suggestion
        parameter_config = next(param for param in pipe.parameters if param["name"] == parameter_name)
        if parameter_config is None:
            assert parameter["expression"] == ""
        else:
            assert parameter["expression"] == parameter_config.default
```

tokkoro (changes requested):
Small tweak for test then I think it's ready to go :)

tests/test_pipeline_conversion.py
memona008: Got the point. Changed it.

tokkoro (approved):
Good work and thanks for putting up for getting all of the details right :)

PR #109, opened 2023-08-16
Add Makefile for development shortcuts

### LINKED ISSUE #108: Add Makefile for better developer experience
Add `make` recipes for building the dev environment, linting, and running tests to reduce manual steps required, and for a more consistent dev experience across repositories.

(See the `valohai-cli` Makefile for reference.)

teroyks:
Resolves #108

Use standard commands (dev, lint, test) for development – reduces the need to check the README for repository-specific commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,16 @@
+.PHONY: help
+help:
+	@echo "check README for usage"
+
+.PHONY: dev
+dev:
+	pip install -e . -r requirements-test.txt pre-commit
+	pre-commit install
+
+.PHONY: lint
+lint:
+	pre-commit run --all-files
+
+.PHONY: test
+test:
+	pytest
--- a/README.md
+++ b/README.md
@@ -54,12 +54,11 @@ print(config.steps['cool step'].command)
 
 ```bash
 # setup development dependencies
-pip install -e . -r requirements-test.txt pre-commit
-pre-commit install
+make dev
 
 # run linting and type checks
-pre-commit run --all-files
+make lint
 
 # run tests
-pytest
+make test
 ```


teroyks:
I was thinking that if we wanted, we could add some extra features to the makefiles, such as:

- optional tests parameter for `make test`
- linting only the changed files

However, this repository is so small that linting and testing all the files is very quick, so perhaps that is more trouble than it is worth.

Makefile
```diff
@@ -0,0 +1,16 @@
...
+.PHONY: dev
+dev:
+	pip install -e . -r requirements-test.txt pre-commit
+	pre-commit install
+
+.PHONY: lint
+lint:
+	pre-commit run --all-files
+
+.PHONE: test
```
ruksi, line 14: :telephone_receiver:

ruksi (changes requested):

Makefile
teroyks: What phone? I see no phone! 🙂 
(pushed a fixup commit)

ruksi (approved):
> I was thinking that if we wanted, we could add some extra features to the makefiles, such as:
> 
> * optional tests parameter for `make test`
> * linting only the changed files
> 
> However, this repository is so small that linting and testing all the files is very quick, so perhaps that is more trouble than it is worth.

yeah, I'd like to make that you could type `make test -- path/to/files` or something which would just run those specific tests / lints but I haven't hand the willpower to check if it would even be possible without `$VARIABLES` :P

for the lint, my personal syntax preference would be `make lint.changed` but dunno what we collectively land on :thinking: 

and probably not worth for this repository :+1:

PR #111, opened 2023-08-17
Pipeline Parameters: use non-None falsy defaults as is in conversion

### LINKED ISSUE #110: Pipeline parameter conversion for falsy values result in empty string
False and 0 should be valid defaults and shouldn't be converted into empty string.

Introduced by #107

tokkoro:
fixes #110 

Falsy parameter defaults were mistakenly cleaned into empty strings by #107


memona008:
LGTM

PR #112, opened 2023-09-04
Validating empty values as int/float gives more specific error

hylje:
The current errors have confusing grammar for empty values. E.g. "is not an integer".

### DIFF
--- a/valohai_yaml/objs/parameter.py
+++ b/valohai_yaml/objs/parameter.py
@@ -113,12 +113,18 @@ class Parameter(Item):
             try:
                 value = int(str(value), 10)
             except ValueError:
-                errors.append(f"{value} is not an integer")
+                if value == "":
+                    errors.append("No value supplied")
+                else:
+                    errors.append(f"{value} is not an integer")
         elif self.type == "float":
             try:
                 value = float(str(value))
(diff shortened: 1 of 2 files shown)


valohai_yaml/objs/parameter.py
```diff
@@ -106,12 +106,18 @@ def _validate_type(self, value: ValueAtomType, errors: List[str]) -> ValueAtomTy
             try:
                 value = int(str(value), 10)
             except ValueError:
-                errors.append(f'{value} is not an integer')
+                if not value:
+                    errors.append('No value supplied')
```
akx, line 110: Hmm... What happens here if you pass in `False` (boolean, not a string)? I think it'll say "No value supplied", but that sounds incorrect...

hylje: Good idea, I've tightened up the logic and added some more tests.

I also made the call to properly complain about validating Nones instead of tossing IndexError. Github actions seem to be feeling bad so let's try those again tomorrow.

akx (changes requested):
Will need a rebase now that #123 was merged, but looks good otherwise

akx (approved):
LGTM (rebased)

PR #113, opened 2023-09-04
Do not keep empty Node `override` in serialized output

akx:
Analogue of 641bc3b89e4518a105c447e36028db432c6aa19a.


euvolen-valohai (approved):
LGTM

PR #114, opened 2023-09-14
Refactor execution resources to a reference

teroyks:
Refs valohai/meta#275

Prepare for reusing the resource definition in the Kubernetes worker configuration to avoid copy/paste – at the moment, these two look identical.

If they later turn out to have different requirements, will then refactor to separate definitions.

### DIFF
--- a/valohai_yaml/schema/endpoint.yaml
+++ b/valohai_yaml/schema/endpoint.yaml
@@ -49,30 +49,11 @@ properties:
     additionalProperties: false
     properties:
       cpu:
-        type: object
-        description: CPU resource requirements and limits in vCPU counts e.g. 0.1 is 10% of one CPU.
-        additionalProperties: false
-        properties:
-          min:
-            type: number
-          max:
-            type: number
+        "$ref": "./workflow-resource-cpu.json"
       memory:
-        type: object
-        description: Memory requirements and limits in MB e.g. 100 is 100 MB.
-        additionalProperties: false
-        properties:
-          min:
-            type: number
-          max:
-            type: number
+        "$ref": "./workflow-resource-memory.json"
       devices:
-        type: object
-        description: "Devices required e.g. 'nvidia.com/gpu: 1' for one NVIDIA GPU."
-        patternProperties:
-          "^.+$":
-            type: number
-            minimum: 1
+        "$ref": "./workflow-resource-devices.json"
 oneOf:
   - required:
       - server-command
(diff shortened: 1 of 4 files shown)


valohai_yaml/schema/endpoint.yaml
```diff
@@ -44,35 +44,7 @@ properties:
...
-          max:
-            type: number
-      devices:
-        type: object
-        description: "Devices required e.g. 'nvidia.com/gpu: 1' for one NVIDIA GPU."
-        patternProperties:
-          "^.+$":
-            type: number
-            minimum: 1
+    "$ref": "./execution-resources.json"
```
ruksi, line 47: would name it something else as it's not always Execution we are talking about, in Deployment Endpoint context, it's not an Execution we are running, but a Deployment Version Endpoint (effectively a Kubernetes Deployment with a number of pods and containers)

maybe just `resources.json` as there is nothing else that name would clash with? Or `workload-resources.json` to lessen potential confusion why something Execution is used in Deployment Endpoints?

valohai_yaml/schema/execution-resources.yaml
```diff
@@ -0,0 +1,29 @@
+type: object
+description: Resource requirements and limits for the endpoint.
```
ruksi, line 2: the description probably either needs to be modified case by case or made more generic for Kubernetes workers?

ruksi (changes requested):

valohai_yaml/schema/endpoint.yaml
teroyks: Workload resources sounds good to me – I started with just `resources` but thought that sounded a bit too generic. Agree that if this is not just for executions, the current name is misleading – will change!

valohai_yaml/schema/execution-resources.yaml
teroyks: In that case, it might make more sence to just generalize the resource properties instead of the whole resources node?

ruksi: I'm a bit pessimistic  that these descriptions are actually read that much TBH so I would just go with a generalized description 🤔 But you can go either way, both approaches are easy to change if need-be 😆

ruksi: both work!

what I would advise against is copy-pasting the definitions, as we probably want parity in the naming of the properties for both use cases to be the same

teroyks: Refactored this to have separate references for all properties.
Still requires naming the properties in each schema though but I don’t think there’s a way to avoid this that would be easier and more clear than just copying the same property names to each schema that uses these.

This allows, however, setting different resource descriptions and using a different combination of properties in different schemas.

ruksi (approved):
lgtm, the best we can get for now I guess :thinking: 

easy to refactor later anyway, now it at least hints to keep similar names with the ref sources

PR #115, opened 2023-09-18
Add optional source-path Step option

hylje:
This is to support changing the Step's source file with `vh yaml step foo.py` invocations, which is implemented later in valohai-utils.

There are no blockers to merge this, as it is a completely optional field.

### DIFF
--- a/tests/test_command.py
+++ b/tests/test_command.py
@@ -34,8 +34,23 @@ def test_command_override(example1_config):
 
 def test_nonexistent_interpolation_keys():
     empty_parameter_map = ParameterMap(parameters={}, values={})
-    interp_command = build_command(['Where are the ${shell_unicorns}? The {parameters} are here!'], empty_parameter_map)
+    interp_command = build_command(
+        ['Where are the ${shell_unicorns}? The {parameters} are here!'],
+        empty_parameter_map,
+    )
     assert interp_command == ['Where are the ${shell_unicorns}? The  are here!']
+
+
+def test_interpolate_special():
+    empty_parameter_map = ParameterMap(parameters={}, values={})
+    interp_command = build_command(
+        ['python {source-path}'],
+        empty_parameter_map,
+        special_interpolations={
+            'source-path': 'foo.py',
+        },
+    )
+    assert interp_command == ['python foo.py']
 
 
 parameter_test_values = {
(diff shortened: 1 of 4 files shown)


valohai_yaml/commands.py
```diff
@@ -44,22 +48,27 @@ def _replace_interpolation(parameter_map: ParameterMap, match: 'Match[str]') ->
...
     :param parameter_map: A ParameterMap object containing parameter knowledge.
+    :param special_interpolations: a str-str dict containing special interpolations
 
     :return: list of commands
     """
     if isinstance(parameter_map, list):
         raise TypeError("Passing in lists as ParameterMaps is no longer supported.")
 
+    special: Dict[str, str]
+    special = {} if special_interpolations is None else special_interpolations
```
akx, line 70: This makes me a bit sad. Isn't mypy placated by 

```suggestion
    if special_interpolations is None:
        special_interpolations = {}
```
?

tests/test_command.py
```diff
@@ -34,7 +34,11 @@ def test_command_override(example1_config):
 
 def test_nonexistent_interpolation_keys():
     empty_parameter_map = ParameterMap(parameters={}, values={})
-    interp_command = build_command(['Where are the ${shell_unicorns}? The {parameters} are here!'], empty_parameter_map)
+    interp_command = build_command(
+        ['Where are the ${shell_unicorns}? The {parameters} are here!'],
+        empty_parameter_map,
+        special_interpolations={}
+    )
```
akx, line 41: Doesn't look like this needed to be changed since `special_interpolations` has a default?

akx (changes requested):
Beyond the inline comments, can you rebase to have Ruff add trailing commas?

Also, I'm always a bit on the edge about adding even more fields to the YAML schema, but...

valohai_yaml/commands.py
hylje: it is not, hence the spell casting

tests/test_command.py
hylje: Not needed, no, the field is optional mostly because it is theoretically a public API but considering it mandatory internally should be ok.

hylje: I dropped it off the old test and added a new test specifically for special interpolations.

PR #118, opened 2023-09-22
More lenient Override parsing

akx:
Refs customer issue trying to look at a legacy config that's unparseable by newer valohai-yamls.


hylje (approved):
lgtm, reasonable changes & test case

PR #119, opened 2023-09-25
Drop support for Python 3.7

akx:
It's been EOL for a while.

Sibling PR to https://github.com/valohai/valohai-cli/pull/289.

PR #120, opened 2023-09-26
Use proper dict type definitions

teroyks:
Related to #119 

Support for Python 3.7 dropped, so proper types can be used instead of comments.


teroyks:
Worked in local testing but failed in CI; maybe take another look later.

PR #121, opened 2023-09-28
Define Kubernetes Workflow Resources

teroyks:
Related to https://github.com/valohai/meta/issues/275

Add a `resources` property to Step for defining Kubernetes resources.

All resources (cpu, memory, devices) are required if `resources` is given.

## Example step definition

``` yaml
- step:
    [...]
    resources:
      cpu:
        min: 0.1
        max: 1
      memory:
        min: 50
        max: 100
      devices:
        nvidia.com/gpu: 1
        nvidia.com/cpu: 2
```

## Testing

- `examples/step-with-resources.yaml` – contains all resource definition properties
- `examples/step-with-partial-resources.yaml` – an example of defining only some of the properties
- `error_examples/step-invalid-resources.yaml` – an example of invalid resources that cause a validation error
- `test_workload_resources.py` – unit tests for the WorkloadResources object

## Open questions

- [x] Ok to require all resources, or no resources at all?
  - do not require all resources, properties may be omitted
- [x] Resources now available to all step definitions – is this ok?
  - yes


teroyks:
@tomi @ruksi Should the resources be marked as Kubernetes-related somehow?

ruksi:
> @tomi @ruksi Should the resources be marked as Kubernetes-related somehow?

I wouldn't mark them as Kubernetes specific, AFAIR we've had other cases where e.g. limiting the resources on executions has been on the table (e.g. a shared instance) and thus e.g. `memory.max` could also be utilized there :thinking:

teroyks:
> Did you have something in mind where we would need to limit their availability?

No, just wanted to make sure there wasn’t some restriction I was unaware of.

teroyks:
Will tidy up the commit history before merging.

valohai_yaml/objs/workload_resources.py
```diff
@@ -0,0 +1,82 @@
...
+    def _parse_args(self, resources: dict) -> None:
+        if not resources:
+            return
+
+        if cpu := resources.get("cpu"):
+            self.cpu = ResourceCPU(cpu)
+        if memory := resources.get("memory"):
+            self.memory = ResourceMemory(memory)
+        if devices := resources.get("devices"):
+            self.devices = ResourceDevices(devices)
```
ruksi, line 37: on all of the `Item` subclasses we have defined here, the interface is a bit off

+ the constructor `__init__` should take arguments in their final form, not a dict blob that you parse, probably the right interface in here would be `cpu: ResourceCPU, memory: ResourceMemory, devices: ResourceDevices`
  + example: https://github.com/valohai/valohai-yaml/blob/master/valohai_yaml/objs/step.py
+ you shouldn't implement custom `_parse_args` but either use the superclass `parse` class method or override it with your own e.g.
  + base: https://github.com/valohai/valohai-yaml/blob/master/valohai_yaml/objs/base.py#L42-L52
  + override: https://github.com/valohai/valohai-yaml/blob/master/valohai_yaml/objs/step.py#L70-L80

you can take more inspiration from most of the other `Item` subclasses defined in the `objs` directory e.g.

+ https://github.com/valohai/valohai-yaml/blob/master/valohai_yaml/objs/endpoint.py
+ https://github.com/valohai/valohai-yaml/blob/master/valohai_yaml/objs/task.py
+ https://github.com/valohai/valohai-yaml/blob/master/valohai_yaml/objs/parameter.py
+ https://github.com/valohai/valohai-yaml/blob/master/valohai_yaml/objs/environment_variable.py (uses superclass `parse` as not defined)

valohai_yaml/objs/workload_resources.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class ResourceCPU(Item):
+    """CPU configuration."""
+
+    # TODO after removing support for Python 3.8 set cpu_resource type: OrderedDict[str, int]
+    def __init__(self, cpu_resource) -> None:  # noqa: ANN001
```
ruksi, line 44: for the typing, probably `Dict[str, Any]` should be right call that we have aliased as `SerializedDict` e.g. https://github.com/valohai/valohai-yaml/blob/master/valohai_yaml/objs/endpoint.py#L36

but as requested previously, `ResourceCPU.__init__` should take simply `min: int, max: int` and a `parse` override should be taking a dictionary for the parsing of serialized objects

ruksi (changes requested):
I think the new objects do not implement the established `Item` interface, see comments

valohai_yaml/objs/workload_resources.py
teroyks: Right, on it.

teroyks: Items refactored.

valohai_yaml/objs/workload_resources.py
teroyks: That took care of the typing! 👍

valohai_yaml/objs/step.py
```diff
@@ -44,6 +45,7 @@ def __init__(
         icon: Optional[str] = None,
         category: Optional[str] = None,
         source_path: Optional[str] = None,
+        resources: Optional[OrderedDict] = None,
```
ruksi, line 48: I think you missed this? this should also take in `WorkloadResouces` and not a serialized dict 😉

valohai_yaml/objs/workload_resources.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __repr__(self) -> str:
+        """Memory data."""
+        return f'ResourceMemory("max": {self.max}, "min": {self.min})'
+
+
+class ResourceDevices(Item):
+    """Devices configuration."""
+
+    def __init__(self, devices: dict) -> None:
```
ruksi, line 42: not sure if `dict` is available, use `Dict` if not 🤔 

the project has `types.py` with various helpers that you can also use

valohai_yaml/objs/workload_resources.py
```diff
@@ -0,0 +1,111 @@
...
+        return ResourceMemory(memory_data.get("max"), memory_data.get("min"))
+
+    @classmethod
+    def _parse_devices(
+        cls,
+        devices_data: Optional[dict],
+    ) -> Optional["ResourceDevices"]:
+        if not devices_data:
+            return None
+        return ResourceDevices(devices_data)
```
ruksi, line 107: this code hints that maybe `ResourceCPU.parse`, `ResourceMemory.parse` and `ResourceDevices.parse` should be implemented? 😆

then the usage could be just:

```
cpu=ResourceCPU.parse(cpu) if cpu else None
```

etc. in the `WorkloadResources.parse`, right?

ruksi (changes requested):
comments + as mentioned, types might need a refresh

valohai_yaml/objs/workload_resources.py
teroyks: Sure – since those are just called from this one class (for now), I thought that is maybe a bit unnecessary and less code would do – but I’ll implement those!

(And who knows, the usage might of course expand later to other places as well.)

teroyks: Turns out using `parse` was less code overall. : )

valohai_yaml/objs/step.py
teroyks: That was because of the custom handling of the resources; after fixing that this now works with the proper type as well; will update.

teroyks: Fixed.

valohai_yaml/objs/workload_resources.py
teroyks: Changed to SerializedDict (which is what actually is passed to the function now).

ruksi (approved):
I think it looks good now, rebase and we can go ahead! :100:

PR #124, opened 2023-10-02
Add upload-store to step

euvolen-valohai:
Solv.: https://github.com/valohai/roi/issues/5438
ref. https://github.com/valohai/roi/pull/5673
- add custom-store filed to allow user to add data store to yaml file

### DIFF
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -110,3 +110,4 @@ def test_error_list():
         ("Additional properties are not allowed" in err) for err in errs
     )  # pragma: no branch
     assert any(("required property" in err) for err in errs)  # pragma: no branch
+    assert any(("0 is not of type 'string'" in err) for err in errs)
--- a/valohai_yaml/objs/step.py
+++ b/valohai_yaml/objs/step.py
@@ -40,6 +40,7 @@ class Step(Item):
         environment_variables: Iterable[EnvironmentVariable] = (),
         environment: Optional[str] = None,
         description: Optional[str] = None,
+        upload_store: Optional[str] = None,
         time_limit: Optional[datetime.timedelta] = None,
         no_output_timeout: Optional[datetime.timedelta] = None,
         icon: Optional[str] = None,
@@ -56,6 +57,7 @@ class Step(Item):
         self.environment = str(environment) if environment else None
         self.icon = str(icon) if icon else None
         self.category = str(category) if category else None
+        self.upload_store = str(upload_store) if upload_store else None
 
         self.outputs = list(outputs)  # TODO: Improve handling
         self.mounts = check_type_and_listify(mounts, Mount)
@@ -81,6 +83,7 @@ class Step(Item):
         )
         kwargs["source_path"] = kwargs.pop("source-path", None)
         kwargs["stop_condition"] = kwargs.pop("stop-condition", None)
+        kwargs["upload_store"] = kwargs.pop("upload-store", None)
         inst = cls(**kwargs)
         inst._original_data = data
         return inst
@@ -116,6 +119,7 @@ class Step(Item):
             ("source-path", self.source_path),
             ("resources", self.resources),
             ("stop-condition", self.stop_condition),
+            ("upload-store", self.upload_store),
         ]:
             serialize_into(
                 val,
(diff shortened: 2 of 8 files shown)


valohai_yaml/schema/step.yaml
```diff
@@ -22,6 +22,9 @@ properties:
   image:
     type: string
     description: The Docker image to use for this step.
+  custom-store:
+    type: string
+    description: The output data store name.
```
akx, line 27: Is it actually the name? Or is it a slug? Or an UUID?

akx (changes requested):
Beyond the inline test: missing tests.

valohai_yaml/schema/step.yaml
euvolen-valohai: It might be both. I will change description

valohai_yaml/objs/step.py
```diff
@@ -39,6 +39,7 @@ def __init__(
         environment_variables: Iterable[EnvironmentVariable] = (),
         environment: Optional[str] = None,
         description: Optional[str] = None,
+        custom_store: Optional[str] = None,
```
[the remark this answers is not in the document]
euvolen-valohai: Considering the fact that execution should not run with invalid store, renaming field to upload-store makes everything much easier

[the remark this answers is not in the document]
euvolen-valohai: Maybe it is even better to use only store name instead of allowing both name and uuid? Store UUID is not easily found on frontend and it is always more chances to may a mistake with it.

valohai_yaml/utils/upload_store.py
```diff
@@ -0,0 +1,16 @@
...
+
+from valohai_yaml.excs import ValidationError
+
+
+def parse_upload_store(upload_store: Optional[str]) -> Optional[str]:
+    if not upload_store:
+        return None
+    import re
+
+    if (isinstance(upload_store, str) and len(upload_store) <= 64) or re.match(
```
akx, line 11: The re.match part of this expression will never be visited for strings that could be valid uuids, since they'll be less than 64 characters long.

valohai_yaml/utils/upload_store.py
```diff
@@ -0,0 +1,16 @@
...
+        r"^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$",
```
akx, line 12: This is currently matched case-sensitively; an UUID would be just as valid written in upper-case.

valohai_yaml/utils/upload_store.py
```diff
@@ -0,0 +1,16 @@
+from typing import Optional
+
+from valohai_yaml.excs import ValidationError
+
+
+def parse_upload_store(upload_store: Optional[str]) -> Optional[str]:
```
akx, line 6: I'm not sure we need this "parsing" (actually validation) in valohai-yaml at all. Do we have precedent for doing this for e. g. the environment field (which is similar in purpose to this new field)?

tests/test_validation.py
```diff
@@ -101,7 +101,7 @@ def test_popo_validation():
 def test_raise():
     with pytest.raises(ValidationErrors) as ei:
         validate(invalid_obj, raise_exc=True)
-    assert list(ei.value)  # test iteration over errors
+    assert len(list(ei.value)) == 5  # test iteration over errors
```
akx, line 104: It would be better to test the actual errors in some degree rather than the magic number 5.

The comment is also no longer accurate.

You could add an explicit test for `"upload-store": 0` rather than piggyback it on here.

akx (changes requested):
Sorry for the short review comments! First ever code review I've done on the Tokyo Metro though...

valohai_yaml/utils/upload_store.py
akx: Why though?

euvolen-valohai: I thought it might be done on roi. But I wrote validation via maxLength in schema validation

akx (approved):
👍 Thanks!

PR #126, opened 2023-10-16
Omit None parameters

tokkoro:
Users should be able to omit parameters fully. Empty string can be a valid value so it cannot be used as a signal for omitting but we need to use None value or some other means.

### DIFF
--- a/tests/test_command.py
+++ b/tests/test_command.py
@@ -100,3 +100,15 @@ def test_parameter_value_with_falsy_values(example1_config):
         ],
     )
     assert command[0] == "env ds='' eps=0 runrunrun"
+
+
+def test_parameter_omit_with_none_value(example1_config):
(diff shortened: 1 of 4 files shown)


tests/test_command.py
```diff
@@ -100,3 +100,15 @@ def test_parameter_value_with_falsy_values(example1_config):
...
+def test_parameter_omit_with_none_value(example1_config):
+    command = example1_config.steps["run training"].build_command(
+        parameter_values={
+            "num-epochs": None,
+        },
+        command=[
+            "echo {parameters}; echo {parameter:num-epochs}",
+        ],
+    )
+    assert "num-epochs" not in command[0]
```
ruksi, line 114: 👍

ruksi (approved):
looks _pristine_ 💎

PR #127, opened 2023-10-16
Add stop-condition field to Step and Task objects/yaml

### LINKED ISSUE #125: Define early stopping rules in the valohai.yaml
It would be great to be able to define early stopping rules in `valohai.yaml` instead of always having to do that in the UI.

hylje:
Resolves #125

The field will be linted as well by trying to parse it.

The field is a simple string value, so it doesn't get its own item/object type. It is part of the Step or Task object, and linting is delegated to a utility function.

### DIFF
--- a/valohai_yaml/utils/lint.py
+++ b/valohai_yaml/utils/lint.py
@@ -1,9 +1,17 @@
-from typing import TYPE_CHECKING, Any, Iterable
+from typing import TYPE_CHECKING, Any, Iterable, Optional
+
+from leval.rewriter_evaluator import RewriterEvaluator
+from leval.universe.verifier import VerifierUniverse
 
 from valohai_yaml.types import LintContext
 
 if TYPE_CHECKING:
     from valohai_yaml.lint import LintResult
+
+
+class _KeywordRewriterEvaluator(RewriterEvaluator):
+    def rewrite_keyword(self, kw: str) -> str:
+        return f"_{kw}"
 
 
 def lint_iterables(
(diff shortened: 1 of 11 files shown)


valohai_yaml/utils/lint.py
```diff
@@ -17,3 +25,27 @@ def lint_iterables(
...
+                break
+        else:
+            return
+        try:
+            evl.evaluate_expression(expression)
+        except Exception as e:
+            lint_result.add_error(
+                f"{object_type.capitalize()} {context[object_type].name}, "
+                f"`{field_name}` is not a valid expression: {e}",
+            )
```
ruksi, line 51: the suggestion is opinionated, use parts that you might more clean, or none at all :stuck_out_tongue: 

mainly declaring things only as they are needed and reducing the nesting

```suggestion
def lint_expression(
    lint_result: "LintResult",
    context: LintContext,
    field_name: str,
    expression: Optional[str],
) -> None:
    if expression is None:
        return

    object_type = None
    for key in context:
        if key in ["task", "pipeline", "step"]:
            object_type = key
            break

    if object_type is None:
        return

    universe = VerifierUniverse()
    evaluator = _KeywordRewriterEvaluator(universe, max_depth=8)
    try:
            evaluator.evaluate_expression(expression)
    except Exception as e:
        lint_result.add_error(
            f"{object_type.capitalize()} {context[object_type].name}, "
            f"`{field_name}` is not a valid expression: {e}",
        )
```

ruksi (changes requested):
> Resolves https://github.com/valohai/valohai-yaml/issues/125

I don't believe reading these is implemented on our private project backend so maybe should be `Refs #...` :thinking: 

but this _does_ add the thing to YAML, it's just not used :stuck_out_tongue: 

otherwise, good work! :100: 

one minor suggestion but the previous works just as well, just opinionated readability improvements, bounce back to review if you feel like this is fine :dog:  :coffee:

valohai_yaml/utils/lint.py
hylje: I touched it up a tad and removed the mystery meat for loop to get the object type (the object should pass it in the context instead)

ruksi (approved):
_very nice_

PR #128, opened 2023-10-17
Add task-presets to serializer to expose via commit with config

axelauvinen:
relates: https://github.com/valohai/roi/issues/5200

Exposing task presets from YAML to API via config serializer

### DIFF
--- a/valohai_yaml/objs/utils.py
+++ b/valohai_yaml/objs/utils.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+from enum import Enum
 from typing import TYPE_CHECKING, Any, Callable, Iterable, List, Optional, Type, TypeVar
 from typing import OrderedDict as OrderedDictType
 
@@ -85,4 +86,7 @@ def serialize_into(
 
 
 def _serialize_if_able(v: Any) -> Any:
+    if isinstance(v, Enum):
+        return v.value
+
     return v.serialize() if hasattr(v, "serialize") else v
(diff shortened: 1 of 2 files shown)


valohai_yaml/objs/utils.py
```diff
@@ -74,6 +75,9 @@ def serialize_into(
     if flatten_dicts and isinstance(value, dict):
         value = list(value.values())
 
+    if issubclass(type(value), Enum) and not hasattr(value, "serialize"):
```
akx, line 78: I think this belongs in `_serialize_if_able`, below, and it should probably be just `if isinstance(value, Enum):`?

akx (changes requested):

valohai_yaml/objs/utils.py
axelauvinen: I'm out of my comfort zone here, can you elaborate a bit how it would belong in the `_serialize_if_able`? :D I wouldn't necessarily bring enum value unwrapping into that function, as then the name should be `_serialize_if_able_or_unwrap_enum`, right?

akx: (We want enums to be serialized as their values, so I think the name is fine.)

axelauvinen: Gotcha, thanks!

axelauvinen: Now fixupped

akx (approved):
Nice!

valohai_yaml/objs/utils.py
ruksi: is a better place for it!

ruksi: or, well, now you can't define custom `serialize` for enums which was the original idea

but the place is better!

PR #130, opened 2023-10-30
Rename step_with_tasks fixture; DRY test

akx:
`step-with-tasks` didn't make sense; this is analogous to `pipeline-example.yaml`.


akx:
Self-merging as mechanical and CI is green.

PR #131, opened 2023-10-30
Kebab-case task properties and make task type validation more lenient

### LINKED ISSUE #129: Normalize Task object keys to be kebab-case instead of snake_case
Everything else in our YAML is kebab-case, but due to a review oversight on my part the Task keys added in #98 are snake_case.

They should be made kebab-case.

Based on analytics, no one has used task definitions in YAML yet (and happily the snake_cased keys are also optional, and so used even less), so we can safely make this change before it's too late.

akx:
* Primarily: Fixes #129 by changing the inadvertently `snake_case`d keys to `kebab-case` to match the rest of our schema. As discussed in that issue, this should be safe to change now, before it's too late.
* Secondarily: Removes the enumeration of task types from the schema; since `TaskType` itself already has lenient parsing for them, we should let that happen.
* Tertiarily: Adds a test to prevent us from adding more underscores into the schema in the future.

### DIFF
--- a/examples/task-example.yaml
+++ b/examples/task-example.yaml
@@ -46,7 +46,7 @@
 - task:
     step: run training
     name: task 4 logspace
-    type: random_search
+    type: RANDOM-SEARCH # will be normalized
     execution-batch-size: 42
     execution-count: 420
     optimization-target-metric: goodness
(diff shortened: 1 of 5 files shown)


euvolen-valohai (approved):
💯!

PR #132, opened 2023-10-30
Task: add remaining fields (maximum queued executions, on child error)

### LINKED ISSUE #90: Define maximum queued executions for Task
It would be useful to be able to define the number of max queued executions in the `valohai.yaml`.

akx:
Fixes #90

### DIFF
--- a/examples/task-example.yaml
+++ b/examples/task-example.yaml
@@ -49,6 +49,8 @@
     type: RANDOM-SEARCH # will be normalized
     execution-batch-size: 42
     execution-count: 420
+    maximum-queued-executions: 17
+    on-child-error: stop_all-AND_error # will be normalized
     optimization-target-metric: goodness
     optimization-target-value: 7.2
     parameters:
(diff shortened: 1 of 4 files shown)


examples/task-example.yaml
```diff
@@ -49,6 +49,8 @@
     type: RANDOM-SEARCH # will be normalized
     execution-batch-size: 42
     execution-count: 420
+    maximum-queued-executions: 17
+    on-child-error: stop_all-AND_error
```
axelauvinen, line 53: Opinion: Is this written intentionally wrong to represent fail-safe working in `cast` method? Ok, but maybe bit weird for an example file 🤔 I would consider adding weirdness into separate test file and merging those into better written example

axelauvinen (approved):
LGTM, opinionated example botherment

examples/task-example.yaml
akx: Maybe I'll add a comment there too, sure :)

axelauvinen (approved):
🚀

PR #133, opened 2023-11-01
Preserve workload resources structure on serialization

teroyks:
Related to https://github.com/valohai/meta/issues/275

Workload resource keys (`cpu`, `memory`, etc.) need to be preserved so that the UI can correctly parse the resource values.

Serialized resources in step before:
```python
{'resources': [{'max': 20, 'min': 10}, {'max': 1.0, 'min': 0.5}, {'big machine': 2}]}
```

After:
```python
{'resources': {'cpu': {'max': 20, 'min': 10}, 'devices': {'big machine': 2}, 'memory': {'max': 1.0, 'min': 0.5}}}
```

The default for most keys (all of them so far) is to flatten them, which is why the test is a bit backwards (`not in [do_not_flatten]` instead of `key in [do_flatten]`).

### DIFF
--- a/tests/test_serialize_step.py
+++ b/tests/test_serialize_step.py
@@ -1,0 +1,16 @@
+def test_serialize_workload_resources(step_with_resources):
+    """Must not flatten workload resource data."""
+    config = step_with_resources
+    resources = config.steps["contains kubernetes resources"].resources
+
+    assert isinstance(resources, dict), "Resources should be defined."
+    assert "cpu" in resources, "Resources should contain data."
+
+
+def test_serialize_partial_resources(step_with_partial_resources):
+    """Serialized data only contains keys found in the config."""
+    config = step_with_partial_resources
+    resources = config.steps["contains partial workload resources"].resources
+
+    assert "min" in resources["cpu"]
+    assert "max" not in resources["cpu"]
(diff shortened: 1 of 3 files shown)


akx (changes requested):
Can you base the test on the (apparently unused!) `step_with_resources` fixture (i.e. `step-with-resources.yaml`) that was added in #121? When using our examples fixtures, we automagically do both direct and round-tripped variants of the test. As I see it, that should surface the bug.

tests/test_serialize_step.py
```diff
@@ -0,0 +1,16 @@
+def test_serialize_workload_resources(step_with_resources):
+    """Must not flatten workload resource data."""
+    config = step_with_resources
+    resources = config.steps["contains kubernetes resources"].resources
+
+    assert isinstance(resources, dict), "Resources should be defined."
+    assert "cpu" in resources, "Resources should contain data."
```
teroyks, line 7: These lines serve more to demonstrate the serialized result than to actually test the process – without the fix, the serialization crashes before reaching here.

teroyks:
> Can you base the test on the (apparently unused!) `step_with_resources` fixture (i.e. `step-with-resources.yaml`) that was added in #121?

Right. Done, pushed a rebased commit (without the unneeded new test data).

akx (approved):
Good stuff!

PR #135, opened 2023-11-01
Add Task parameter-sets

### LINKED ISSUE #134: Task presets: allow configuring parameter sets for Manual Search
It should be possible to define also a manual search type Task in the `valohai.yaml` with the predefined parameter sets. 

Related discussion: https://valohai.slack.com/archives/C02B7PSLQ3B/p1698836322106279

akx:
Fixes #134

### DIFF
--- a/examples/task-example.yaml
+++ b/examples/task-example.yaml
@@ -77,3 +77,21 @@
           base: 10
           distribution: uniform
           integerify: False
+- task:
+    step: run training
+    name: parameter sets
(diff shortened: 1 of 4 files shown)


hylje (approved):
example, test, parsing

ruksi (approved):
lgtm

could lint that the parameters in parameter sets actually exist

but then again, we _could_ have future use for some arbitrary values in parameter sets :thinking:

PR #142, opened 2024-02-20
Add edge-merge-mode property in Input

memona008:
Refs [#6215](https://github.com/valohai/roi/issues/6215)

New property is introduced in Inputs

**Replace input's default value when edge is overriding the input** ( `edge-merge-mode: replace` ) or Pipeline node has `override` for the input 
**Combine the input's default value with edge's output** (`edge-merge-mode: append`) and Pipeline node does not have `override` for the input

By default each input will have `edge-override-mode: replace`


valohai_yaml/schema/input-item.yaml
```diff
@@ -38,3 +38,6 @@ properties:
           - "suffix"
     default: false
     description: Whether to retain directories when using wildcards for this input.
+  overridable:
+    type: boolean
+    description: Whether to replace or append the input when something else is overriding it.
```
akx, line 43: I think the description is at odds with the name of the property here, and since we'll be stuck forever with what we decide here, it'd better be good...

```suggestion
  override-mode:
    type: string
    enum:
      - replace
      - append
    default: append
    description: Whether to replace the values set for this input, or to append to them when overridden by e.g. a pipeline edge.
```
(Is the description correct?)

memona008: yes, and `default: append ` wouldn't be effective when pipeline node explicitly define `override`. Linter will throw error that append-able input and explicitly overriding inputs with edge are mutually exclusive.
As explicit override will definitely replace the input.

valohai_yaml/pipelines/conversion.py
```diff
@@ -105,3 +107,16 @@ def convert_executionlike_node(
         }
 
         return node_data
+
+    def validate_overriding(self, override: Override, all_inputs: dict) -> None:
+        strictly_non_overridable_inputs = {
+            _input["name"]: _input
```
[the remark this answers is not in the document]
memona008: yeah i was trying to avoid the built-in names as it becomes the shadow name

akx (commented):
So... why would something that has an "append" override mode be called "strictly non-replaceable"? 🤔

memona008:
> So... why would something that has an "append" override mode be called "strictly non-replaceable"? 🤔

If we are seeing it from pipeline node override perspective then the `append` mode would depict that the **default value** is not **replaceable**, however `node: override: inputs` property is designed to `replace` it 
Hence strictly non-replaceable 🥇 :D

valohai_yaml/enums.py
```diff
@@ -0,0 +1,8 @@
+from enum import Enum
+
+
+class InputOverrideMode(Enum):
+    """Input override mode."""
+
+    REPLACE = "replace"
+    APPEND = "append"
```
akx, line 8: This should be in `valohai_yaml.objs.input` like the other input-related enum.

valohai_yaml/objs/input.py
```diff
@@ -37,13 +37,15 @@ def __init__(
         description: Optional[str] = None,
         keep_directories: KeepDirectoriesValue = False,
         filename: Optional[str] = None,
+        override_mode: Optional[bool] = None,
```
akx, line 40: I don't think this type annotation is correct. The default should be the current legacy default.

valohai_yaml/objs/pipelines/execution_node.py
```diff
@@ -41,12 +42,20 @@ def lint(self, lint_result: LintResult, context: LintContext) -> None:
         if self.override is not None:
             self.override.lint(lint_result, context)
             step_inputs = step.inputs.keys()
+            strictly_non_replaceable_inputs = [
+                inp.name for inp in step.inputs.values() if inp.override_mode == InputOverrideMode.APPEND.value
+            ]
```
akx, line 47: I'm not sure this makes sense, _if_ the end goal is that pipeline edges can either replace or append inputs?

akx (changes requested):
I'm not sure about the "strictly overridable" logic here..? I mean, if the end goal is that the user can decide whether inputs from edges will replace or append inputs specified in the node's execution template, why would something be strictly non-overridable..?

valohai_yaml/objs/pipelines/execution_node.py
memona008: if someone choose append here 
and one pipeline edge defines explicit override (which means it would replace) 
its the conflicting thing, how should we tackle this then 🤔 or we should ignore it assuming replace has upper hand

valohai_yaml/objs/input.py
```diff
@@ -37,13 +44,15 @@ def __init__(
         description: Optional[str] = None,
         keep_directories: KeepDirectoriesValue = False,
         filename: Optional[str] = None,
+        override_mode: Optional[InputOverrideMode] = InputOverrideMode.APPEND,
```
akx, line 47: `Optional` implies that `None` is a valid `override_mode`. Is that correct?

valohai_yaml/objs/input.py
```diff
@@ -37,13 +44,15 @@ def __init__(
...
     ) -> None:
         self.name = name
         self.default = default  # may be None, a string or a list of strings
         self.optional = bool(optional)
         self.description = description
         self.keep_directories = KeepDirectories.cast(keep_directories)
         self.filename = filename
+        self.override_mode = override_mode
```
akx, line 55: This should probably do `OverrideMode.cast(override_mode)` like `keep_directories` above (so you'd need to implement `OverrideMode.cast` too).

akx (changes requested):
Beyond the inline comments: tests, please?

valohai_yaml/objs/input.py
memona008: One doesn't have to forcefully define override mode on each input. But as we are already assigning a default value i think Optional can be omitted now.

akx:
cc @noorula, do you think this addresses the issue? ✨

tokkoro (dismissed):
Good stuff. It resolves the issue we are having with customers.

memona008:
Related discussion here: 
https://valohai.slack.com/archives/CML8Z1WTW/p1709727038858249
@drazendee @noorula please ping when do you guys want this change to be merged.

valohai_yaml/schema/input-item.yaml
```diff
@@ -38,3 +38,10 @@ properties:
           - "suffix"
     default: false
     description: Whether to retain directories when using wildcards for this input.
+  edge-merge-mode:
+    type: string
+    enum:
+      - replace
+      - append
+    default: append
```
akx, line 46: The `default` here seems at odds with the Input object's `EdgeMergeMode.REPLACE`.

valohai_yaml/objs/input.py
```diff
@@ -52,3 +70,7 @@ def get_data(self) -> SerializedDict:
         else:
             data.pop("keep_directories", None)
         return data
+
+    def apply_edge_merge_mode(self) -> None:
+        if self.edge_merge_mode == EdgeMergeMode.REPLACE:
+            self.default = []
```
akx, line 76: I'm not sure this is a great idea. I'd expect the configuration object to be immutable.

Feels like the `Override.merge_with_step()` function should know about this instead.

akx (dismissed):

akx:
I disabled auto-merge so an inadvertent approval won't accidentally mainline this before its time.

valohai_yaml/objs/input.py
memona008: I have a strong opinion about it being included in Input as the property belongs to Input and should be applied 
before merging override with step. Mutating default based on property is not doing any damage as it's being serialized for node data right away which is required in Pipeline conversion.

If we still presist to do it in merge_with_step 
There we still have to mutate defaults before merge_dicts is called 

or if we do it in merge_dicts it becomes messy as that method is working for any type and this edge-merge-mode is specifically related to Input only.

akx: > as it's being serialized for node data right away

_here_, yes. 

If you have a strong opinion about it being included in this class, then I think it should at least be a function that doesn't internally mutate things (i.e. returns a shallow copy with `default` cleared):

```
input = input.with_edge_merge_mode_applied()
```

valohai_yaml/pipelines/conversion.py
```diff
@@ -91,6 +91,9 @@ def convert_executionlike_node(
                 step_data.pop("no-output-timeout"),
             )
 
+        for key, input in step.inputs.items():
+            step.inputs[key] = input.with_edge_merge_mode_applied()
```
akx, line 95: 1. Should this occur before `step.serialize()` above?
2. Now the `step` is mutated 😁 IOW, if there is a pipeline that has multiple nodes that use the same step, the `convert_executionlike_node()` calls may use a slightly different step. If I'm reasoning about this right, it shouldn't make a difference here, but it's still not good form. Consider:
   ```python
   config = Config()
   print(config.nodes["train"].default)  # ["foo", "bar", "baz"]
   p = PipelineConverter(config, "foo")
   x = p.convert_pipeline(config.pipelines["pipe"])
   print(config.nodes["train"].default)  # []
   ```

valohai_yaml/objs/input.py
```diff
@@ -1,10 +1,27 @@
...
+
+    REPLACE = "replace"
+    APPEND = "append"
+
+    @classmethod
+    def cast(cls, value: EdgeMergeModeValue) -> "EdgeMergeMode":
+        if isinstance(value, EdgeMergeMode):
+            return value
+        if not value:
+            return EdgeMergeMode.APPEND
```
tokkoro, line 23: The default behavior should be replacing. Is this for the backward compatibility?

valohai_yaml/pipelines/conversion.py
```diff
@@ -111,6 +113,11 @@ def convert_executionlike_node(
                 step_data.pop("no-output-timeout"),
             )
 
+        # Shallow-copy the step before we mutate it,
+        # so the changes don't reflect into `self.config`
+        step = copy.copy(step)
+        step.inputs = OrderedDict((key, input.with_edge_merge_mode_applied()) for key, input in step.inputs.items())
+
```
tokkoro, line 120: I think we should override inputs only when there is an actual edge configured. If no edge is configured to override this specific input it should use the defaults. If I read this correctly it's also clearing the defaults from inputs that are not part of an edge.

tokkoro (changes requested):
Possible just a confusion on my part, but I want to clarify them before releasing this.

valohai_yaml/pipelines/conversion.py
tokkoro: Based on this and @akx and my discussion today. This feature cannot be done in valohai-yaml as a part of the logic depends on if there is an actual edge(s) and thus should be handled when the actual incoming edge data is resolved in the system actually managing the pipeline (Roi for example). Also the configuration can be altered after it's read from the config file and before it's send to a server.

tokkoro:
work continues in #153

PR #146, opened 2024-03-05
🔨 Use `uv` for handling packages if available

teroyks:
If the `uv` command is found, uses `uv pip` to install the dependencies.

Standard `pip` still used as a fallback.

If `uv` is available, `make dev` will use that even if the virtual environment was originally created with `virtualenv` (seemed to work just fine).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,17 @@
+UV := $(shell command -v uv)
+ifdef UV
+	PIP_INSTALL = uv pip install
+else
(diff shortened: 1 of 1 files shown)


teroyks:
Tested building the environment with `uv`. Seems to work even when converting from the original environment created with `virtualenv` – but maybe some additional testing and fixing needed for different environments.

akx (approved):

PR #147, opened 2024-03-06
Support list as pipeline parameter expression

memona008:
Resolves [300](https://github.com/valohai/valohai-cli/issues/300)
## Changes in this PR:
- Added convert expression function to convert YAML lists as list expressions, which behaves the same as pipelines on the web application
- Add pipeline parameters support for PipelineConverter

### DIFF
--- a/tests/test_pipeline_conversion.py
+++ b/tests/test_pipeline_conversion.py
@@ -81,4 +81,30 @@ def test_pipeline_parameter_conversion(pipeline_with_parameters_config):
                 }
             else:
                 assert parameter["expression"] == expression_value
-                assert type(parameter["expression"]) == type(expression_value)
+                assert type(parameter["expression"]) is type(expression_value)
+
+
+def test_pipeline_parameter_conversion_with_args(pipeline_with_parameters_config):
+    pipeline = pipeline_with_parameters_config.pipelines["Example Pipeline with Parameters"]
+    args = {
+        "param0-float": 1.5,
+        "param0-int": 3,
+        "param0-string": "iceberg",
+        "param0-flag": True,
+        "listings": ["msg", "pfa"],
+    }
+    result = PipelineConverter(
+        config=pipeline_with_parameters_config,
+        commit_identifier="latest",
+        parameter_arguments=args,
+    ).convert_pipeline(pipeline)
+    params = result["parameters"]
+
+    assert params["param0-float"]["expression"] == 1.5
+    assert params["param0-int"]["expression"] == 3
+    assert params["param0-string"]["expression"] == "iceberg"
+    assert params["param0-flag"]["expression"]
+    assert params["listings"]["expression"] == {
+        "style": "single",
+        "rules": {"value": ["msg", "pfa"]},
+    }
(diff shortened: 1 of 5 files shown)


valohai_yaml/pipelines/conversion.py
```diff
@@ -102,3 +103,8 @@ def convert_executionlike_node(
         }
 
         return node_data
+
+    def convert_expression(self, expression: ExpressionValue) -> ExpressionValue:
+        if isinstance(expression, list):
+            return ",".join(expression)
```
akx, line 109: Shouldn't this take `multiple-separator` into account?

memona008: my bad, it should yes

memona008: Related discussion: https://valohai.slack.com/archives/C3J44LTL6/p1710758618962599

hylje:
Adopting this branch and PR with a slightly different approach

valohai_yaml/pipelines/conversion.py
```diff
@@ -14,6 +14,16 @@
...
 
 
+class VariantExpression(TypedDict):
+    """Variant expression template."""
+
+    style: str
+    rules: Dict[str, Any]
+
+
+ExpressionValue = Union[str, int, bool, float, VariantExpression]
```
dangquangdon, line 24: How would this `VariantExpression` dictionary affect they way roi and/or `valohai-utils` parsing the value now? Btw, should we call it `ExpressionValueVariant`?

hylje: Backend accepts variant expressions as parameter values, and it is the only way it accepts list parameters. It works for all parameter value types though.

Valohai-utils is not involved in this process at all. We pass a list in structured format to backend, and backend receives a list in structured format. If you pass a list parameter as a command line parameter and parse command line parameters in a Python script using `valohai-utils`, it must use comma separated values as that is the only format `valohai-utils` accepts. Other scripts or programs can accept different parameter passing styles, which is why we support custom multiple-separators.

You can also use file based configuration to read the structured list parameter as JSON (`valohai/config/parameters.json`), which skips the whole parsing step. This is what `valohai-utils` automatically does for parameters that haven't been overridden as command line parameters.

tests/test_pipeline_conversion.py
```diff
@@ -73,5 +73,38 @@ def test_pipeline_parameter_conversion(pipeline_with_parameters_config):
...
+
+def test_pipeline_parameter_conversion_with_args(pipeline_with_parameters_config):
+    pipeline = pipeline_with_parameters_config.pipelines["Example Pipeline with Parameters"]
+    args = {
+        "param0-float": 1.5,
+        "param0-int": 3,
+        "param0-string": "iceberg",
+        "param0-flag": True,
+        "listings": ["msg", "pfa"],
+    }
```
dangquangdon, line 95: where does this come from in practice?

dangquangdon (approved):

tests/test_pipeline_conversion.py
hylje: I added pipeline argument parsing when converting for a future CLI enhancement. This PR lets you send lists as pipeline parameters if you make the values you want default in YAML config.

PR #149, opened 2024-03-13
Parameter categories

akx:
This PR adds YAML support for categorizing parameters and pipeline parameters. An empty string will be considered a `null` category.

While doing that, it also adds `description` to pipeline parameters, for parity with regular parameters.

### DIFF
--- a/examples/example1.yaml
+++ b/examples/example1.yaml
@@ -35,11 +35,13 @@
         description: How many supervised samples are used
         type: integer
         default: 60000
+        category: Samples
 
       - name: unlabeled-samples
         description: How many unsupervised samples are used
         type: integer
         default: 60000
+        category: Samples
 
       - name: encoder-layers
         description: List of layers for f
@@ -61,10 +63,12 @@
         type: string
         default: SELECT * FROM mytable
         widget: sql
+        category: Database
 
       - name: output-alias
         type: string
         default: my-alias
+        category: Output
         widget:
           type: DatumAlias
           settings:
--- a/examples/pipeline-with-parameters-example.yaml
+++ b/examples/pipeline-with-parameters-example.yaml
@@ -25,6 +25,7 @@
           - train_parallel.parameter.id
         default: 123
         category: Apples
+        description: The ID of the thing to train
       - name: param0-float
         targets:
           - train.parameter.param0-float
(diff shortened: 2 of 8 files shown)


tokkoro:
This was my first thought as well. We discussed it with @ruksi and he suggested that we could nest parameters under categories YAML objects. The benefit from that would be that the definition would then be only couple of lines more per category and not one line more per parameter. But this would require the reader to first check if the content of `step.parameters` is parameters or categories and possible sub categories would add more indentation on every level.

@akx Do you see an easy way to implement it in the nested manner or should we go by what you have suggested here?

akx:
@tokkoro If you mean

```yaml
parameters:
  Apples:
    foo:
  Oranges:
    bar:
```
then I'm not sure that's a good idea, because it will be ambiguous to parse:
```yaml
parameters:
    name:
        default:
```
could either be a category-less parameter named `name` with an empty `default` or it can be a parameter named `default` under the category `name`.

ruksi:
I was thinking more on the lines of:

```yaml
# ungrouped
- step:
    # ...
    parameters:
      - name: iterations
        type: integer
      - name: rate
        type: float
      - name: email
        type: string
      - name: extra
        type: string

# grouped
- step:
    # ...
    parameters:
      - learning:
        - name: iterations
          type: integer
        - name: rate
          type: float
      - reporting:
        - name: email
          type: string
      - name: extra
        type: string
```

and then you can access them `parameters.learning.iterations` etc.

ruksi:
didn't really think much about the ambiguity but you can get around anything with some naming conventions~

akx:
Naming convention changes are also more backward-incompatible.

ruksi:
> Naming convention changes are also more backward-incompatible.

aka. _very not nice_

ruksi:
I don't have a strong preference in this

I know that **I** would like to copy paste a group name to multiple places :sweat_smile:

ruksi:
then you could have pre-defined parameter groups if you have similar parameters in multiple steps:

```yaml
- parameter-group: learning
  parameters:
    - name: iterations
      type: integer
    - name: rate
      type: float

- step:
    # ...
    parameters:
      - learning: learning
      - reporting:
        - name: email
          type: string
      - name: extra
        type: string
```

ruksi:
> Also backward-incompatible 😁 But you can do that to some degree with YAML anchors and references already.

yeah, I would need to play around with the YAML a bit to make some **real** proposals, this is more for _inspiration_ :sparkles:

PR #152, opened 2024-05-30
Added `reuse-executions` property to pipeline schema

dangquangdon:


valohai_yaml/schema/pipeline.yaml
```diff
@@ -38,6 +38,13 @@ properties:
           required:
             - source
             - target
+  allow-reuse:
```
tarekoraby, line 41: Would it make sense to call it `node-reuse` with true/false values?

tarekoraby: Or `step-reuse`

akx (dismissed):
1. Tests fail.
2. The .gitignore changes seem unrelated?
3. What are the semantics here – is this just the default for the pipeline API `allow_reuse` flag (which, afaiu, only affects reusing executions into execution nodes?), or something else? I can't seem to find a ref to any sort of spec for this.

dangquangdon:
> 3\. What are the semantics here – is this just the default for the pipeline API `allow_reuse` flag (which, afaiu, only affects reusing executions into execution nodes?), or something else? I can't seem to find a ref to any sort of spec for this.

The idea is that if user specify this in `valohai.yaml`, the Automatically reuse switch button in the UI will also reflect it

valohai_yaml/schema/pipeline.yaml
dangquangdon: we called it `allow_reuse` in roi's code too so we could keep it for consistency

tarekoraby: > we called it allow_reuse in roi's code 

Do you mean in the internal code or in the user-facing UI?

What we call it internally doesn't matter to the end users. In the UI we had some preamble text explaining what the feature does and how it works. In `valohai.yaml`, the terminology we use should be self-explanatory, and `allow-reuse` isn't descriptive enough of what the feature does.

valohai_yaml/schema/pipeline.yaml
```diff
@@ -38,6 +38,13 @@ properties:
...
+    type: boolean
+    default: false
+    description: |
+      Set to true to allow Valohai to automatically detect if the executions in
+      steps' nodes could be skipped and reuse the result from a previously successfully
+      run node.
```
tarekoraby, line 47: Somewhere in the description, we should mention the conditions that lead to execution being matched (i.e., that the execution is in the same project, that it has the same inputs, parameters, code commit, and data).

hylje:
The snapshots have changed due to a change in a downstream library version.

I've fixed the snapshot issue in https://github.com/valohai/valohai-yaml/pull/147 which is waiting for review.

valohai_yaml/schema/pipeline.yaml
dangquangdon: sure, let's use `node-reuse`/`step-reuse` as you suggested. However, IMO, from user's perspective, it doens't matter that much because whichever keyword we use, it wouldn't be self-explanatory enough for this feature, user will have to look for the description and/or documentation anyway to be able to understand what the keyword does.

valohai_yaml/schema/pipeline.yaml
dangquangdon: Could you suggest some text for it? Tbh I'm not good with the wording 😂

dangquangdon:
> The snapshots have changed due to a change in a downstream library version that changed sort order.
> 
> I've fixed the snapshot issue in #147 which is waiting for review.

Nice, thanks for the info. Did it fail also in your local machine when you were fixing it? Because it didn't happen in mine 🤔

hylje:
It didn't fail at first but when I upgraded the dependencies it started failing the same way it did in CI (which always installs freshest dependencies)

dangquangdon:
> It didn't fail at first but when I upgraded the dependencies it started failing the same way it did in CI (which always installs freshest dependencies)

I checked the dependencies versions as well and they were the same as in CI, but after your comment, I remove the whole `venv` and create a fresh new one and was able to reproduce it. Seemed like it was some kind of caching issue for me.

valohai_yaml/schema/pipeline.yaml
dangquangdon: added some more text to it 👍

valohai_yaml/schema/pipeline.yaml
akx: TBH, I liked `allow-reuse` more. `node-reuse` is inaccurate, since it does not reuse nodes, but executions that would be executed by some nodes. `reuse-executions` smells like it would be most accurate for what we have now.

akx (commented):
LGTM (not approve-gating yet), if @tarekoraby is OK with the new name.

tarekoraby:
`reuse-executions` LGTM 👍

akx (approved):

PR #155, opened 2024-10-01
Add tests that ensure parameter & input orders are retained

akx:


ruksi:
do we need to consider ourselves with the parameter groups; sorting of those and inside of those?

akx:
@ruksi Parameter categories are just a string on a Parameter object, they don't have an intrinsic ordering (and as is, can't have one).

ruksi:
so a strong **no**

ruksi (approved):
lgtm!

PR #157, opened 2024-11-29
Implement new step.inputs[].download configuration

hylje:
This supports marking specific inputs as either always downloaded or on-demand downloaded. The default is to always download, maintaining backwards compatibility.


akx (approved):
Seems legit. 👍 for the schema doc explaining the idea!

PR #158, opened 2025-01-03
Ensure shebang lines are passed through as expected

akx:
This PR ensures shebang lines remain as-is when commands are joined together.

Other comments won't be handled correctly by `join_commands` yet, though.


hylje (approved):
asserts expectations, shells can now be banged

PR #159, opened 2025-01-15
Elide default reuse-executions and edge-merge-mode values from serialization results

akx:
Keeps the serialized data more minimal.


hylje (approved):
seems all right with tests included

PR #160, opened 2025-02-10
introduce environment_variable_groups to Step class

Kenneth-Schroeder:
I want to make the environment_variable_groups setting accessible from the CLI and from the valohai yaml.
To achieve this, I have created a PR in valohai-cli as well ([valohai-cli#323](https://github.com/valohai/valohai-cli/pull/323)).

I could test my changes on valohai-cli and the local linting here on valohai-yaml, however, the API endpoint seems to validate with an older valohai-yaml version as well, causing an API validation error.


akx (changes requested):

Kenneth-Schroeder:
@akx Anything else we need to address?

akx:
> @akx Anything else we need to address?

Sorry for the delay! Added a test and an example here 👍

PR #161, opened 2025-03-25
Renovations, 2025-03 (drop Python 3.8 support)

akx:
* General tool updates
* Dropping support for Python 3.8 (EOL for almost half a year now)


ruksi (approved):
upgraded linters and checkers

applying the suggested fixes

lgtm! :+1:

PR #163, opened 2025-05-09
✅ Sort selected test snapshot lines

teroyks:
`test_validation` kept failing for me locally – the error messages are otherwise identical, but they appeared in a different order than in the snapshot.

Since the content of the validation error messages is what is important, it should be enough to just test that both contents have the same lines, no matter which order the messages are in.

Add optional sorting of the validation error results, so the order of the messages does not matter any more.
There are no functional changes to the validation tests in this PR.

The sorting could maybe have been made mandatory, but since the flakiness has only appeared in one of the tests, make it optional to fix just that test and leave the rest be. This can be changed if flakiness is found in other tests later.

The downside of this change is that the snapshot file is less human-readable. But I doubt anyone regularly reads those.

Snapshot before:
```
  >>> step-missing-required-properties.yaml
  error:   Type validation on step.image.type: True is not of type 'string' (0.step.image)
  ------------------------------------------------------------
  error:   Oneof validation on step.command.oneOf: 8 is not valid under any of the given schemas (1.step.command)
  ------------------------------------------------------------
  error:   Required validation on step.required: 'command' is a required property (0.step)
  ------------------------------------------------------------
  error:   Required validation on step.required: 'name' is a required property (0.step)
  ------------------------------------------------------------
  *** 4 errors, 0 warnings
```

Local error before:
```
  >>> step-missing-required-properties.yaml
  error:   Oneof validation on step.command.oneOf: 8 is not valid under any of the given schemas (1.step.command)
  ------------------------------------------------------------
  error:   Type validation on step.image.type: True is not of type 'string' (0.step.image)
  ------------------------------------------------------------
  error:   Required validation on step.required: 'command' is a required property (0.step)
  ------------------------------------------------------------
  error:   Required validation on step.required: 'name' is a required property (0.step)
  ------------------------------------------------------------
  *** 4 errors, 0 warnings
```

Snapshot and local test output after, with lines for a test sorted:
```
  *** 4 errors, 0 warnings
  ------------------------------------------------------------
  ------------------------------------------------------------
  ------------------------------------------------------------
  ------------------------------------------------------------
  >>> step-missing-required-properties.yaml
  error:   Oneof validation on step.command.oneOf: 8 is not valid under any of the given schemas (1.step.command)
  error:   Required validation on step.required: 'command' is a required property (0.step)
  error:   Required validation on step.required: 'name' is a required property (0.step)
  error:   Type validation on step.image.type: True is not of type 'string' (0.step.image)
```

### DIFF
--- a/tests/__snapshots__/test_validation.ambr
+++ b/tests/__snapshots__/test_validation.ambr
@@ -1,167 +1,152 @@
 # serializer version: 1
 # name: test_bad_examples_cli[endpoint-names-invalid.yaml]
   '''
-  >>> endpoint-names-invalid.yaml
-  error:   Pattern validation on endpoint.name.pattern: 'Server-endpoint' does not match '^[a-z][a-z0-9-]+$' (0.endpoint.name)
-  ------------------------------------------------------------
-  error:   Pattern validation on endpoint.name.pattern: '3-server-endpoint' does not match '^[a-z][a-z0-9-]+$' (1.endpoint.name)
-  ------------------------------------------------------------
-  error:   Pattern validation on endpoint.name.pattern: 'server@endpoint' does not match '^[a-z][a-z0-9-]+$' (2.endpoint.name)
-  ------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


akx (approved):
see it, say it, sorted

PR #166, opened 2025-05-22
🙈 Move a bad warning example from examples to warning examples

ruksi:
don't promote optional flags under "examples" as the filename or contents don't communicate that it is a warning example

... actually I was doing something else and the relative `./` usage of the file was causing annoyance for me, and noticed that it's a wolf in sheep's clothes (a warning example inside our demo examples)

while I was at it, also added comments on the warning examples to make it more obvious they are **bad** examples


teroyks (approved):
Bad example! Bad! Off to your place you go!

PR #168, opened 2025-05-26
Update version number to 0.47.0

teroyks:
Will fix the release tag after this is merged…
Nevermind, will create a new version to get the release properly pushed.


teroyks:
Will not be used; `v0.47.1` fixes version number issue in `__init__`.

Issue #150, opened 2024-03-15
`valohai.yaml` Autocompletion: Schemas & Instructions


teroyks:
Sort out things needed for using autocompletion when editing `valohai.yaml`.

- Demo was presented at Skunk week (@teroyks)
- Convert schema YAML files to JSON (Tero has a Makefile update for it)
- Decide on where to put the files
- Decide on how to update the files
- Write instructions for users
  - VS Code is sufficient as a first step
  - Link to the published base schema (either in user/workspace settings or directly in the file)

PR #170, opened 2025-05-27
Inline JSON Schema in code

akx:
WDYT @teroyks? ✨ 

Refs #150.


akx:
> I like it! I don’t think the refs are quite correct, though – at least VS Code has issues with them. It wants `$ref` → URL to point to a different file with that URL.

Hm, the `jsonschema` library in turn had issues with `"$ref": "#/$defs/..."`, and I don't blame it because that'd mean there'd be a nested `/$defs` at that point in the schema. On the other hand, `/$defs/step` also made it choke.

I guess we'll need to find a scheme that works for both, or alternately have the new export command fix up the references in various styles, which isn't an issue either.

teroyks:
> Hm, the `jsonschema` library in turn had issues with `"$ref": "#/$defs/..."`

That’s weird – it’s directly from the [JSON Schema docs](https://json-schema.org/understanding-json-schema/structuring#defs).

akx:
Yes, that works in the top level schema.

But AIUI, if you have a schema in $defs, # refs to that schema's root, and we don't have $defs in the schemas already nested in $defs...

(Oops, I somehow managed to edit your comment instead of replying... long day 😵)

teroyks:
> But AIUI, if you have a schema in $defs, # refs to that schema's root, and we don't have $defs in the schemas already nested in $defs...

…almost just edited your comment as well. Time to call it a day, I guess. : ) 

Ok, from later in the same doc: “a `$ref` referring to another `$ref` – – is explicitly disallowed”. 
Some more thinking needed, probably.

teroyks (approved):
Is good! 
We can keep on thinking about the exact schema format and update later.

PR #171, opened 2025-06-12
♊️⚠️⬇️ Demote duplicate top-level entity name error to warning

ruksi:
there were multiple users that had duplicate names in their config files which caused our reading of their config files to halt (config is invalid, we can't use it)

this raising of an error caused various issues; issues which will continue even if users remove the duplicates as we'll have to be able to parse historical config files :sweat:

demote to warning; multiple-top-level-entity-names will still continue be "undefined behavior" and we don't guarantee which one we'll be using if there are multiple entities with the same name so the warning is warranted and duplicates removed


teroyks:
This could alternatively be done by just reverting the latest few comments, as the first version was just warnings, and it was only updated to an error after change requested at code review…

noorula:
Were the warnings shown to the customer somewhere by the way? 🤔

ruksi:
> I think it would be a good thing to show it both when linting valohai.yaml and also in the UI when fetching the repo. 🤔

yup, I agree

when fetching the repo as extra messages

image

and maybe

image

image

ruksi:
> Sure. Just quite unfortunate that there are so many noncompliant YAMLs out there!

yeah, it was unfortunate

one thing we can do, like discussed above, is to show the lint warnings on the UI

found [the issue for showing YAML lint errors](https://github.com/valohai/roi/issues/4611), I'll umm link or paste the screenshots there for suggestions where the warnings could be shown

PR #173, opened 2025-08-20
☁️➕ Allow configuring deployments for auto-create

ruksi:
the first draft how one could define deployment configuration so we can create it automatically on pipeline deployment node

I was first thinking of including it under `pipeline.nodes.<nodename>.deployment` but noticed that:

1. the pipeline definition is already quote a bit of nesting and this would make it go one or two levels deeper
2. you might want to auto-create the same deployment from multiple pipelines or even deployment nodes in the same pipeline
    - (expectedly) you'd copy-paste the definition in all places; this can get out-of-sync
    - (if one knowns YAML) you'd use `&` anchors and `*` aliases

# Proposal

you define a top-level `deployment` which holds the default values to use on deployment auto-create:

```yaml
- deployment:
    name: us-west-qa  # name of the deployment both for matching and creating
    defaults:
      target: acme-corp-us-west-cluster  # UUID or the dash-text-indentifier
      allow-cors: True|False  # optional, use platform default if not given
      allow-configuration-snippets: True|False  # optional, use platform default if not given
      # leaves room for further configuration
```

I opted to expose only `target`, `allow-cors` and `allow-*-snippets` as those are the configuration a user can do on the normal interfaces. It leaves room for other more advanced configuration if needed: kubernetes namespace, ingress name and manifest mods.

the values are under `defaults` to both hint that "these won't be applied retroactively" and leaves room for such configuration that _would_ be retroactive, if such a need raises

then you refer to this deployment from a pipeline deployment node:

```yaml
- pipeline:
    [...]
    nodes:
      [...]
      - name: deploy
        type: deployment
        deployment: us-west-qa  # matching for existing ones, or for auto-creation config lookup
        endpoints: [predict-digit]
```

the idea would be that the deployment is created only when required i.e. when a pipeline deployment node would be ran

when creating a pipeline that would refer to such deployment definition, you can still do the normal flow of selecting a deployment to deploy into manually

### DIFF
--- a/valohai_yaml/objs/deployment.py
+++ b/valohai_yaml/objs/deployment.py
@@ -1,7 +1,8 @@
 from typing import Optional
 
+from valohai_yaml.lint import LintResult
 from valohai_yaml.objs.base import Item
-from valohai_yaml.types import DeploymentDefaultsDict
+from valohai_yaml.types import DeploymentDefaultsDict, LintContext
 
 
 class Deployment(Item):
@@ -15,3 +16,12 @@ class Deployment(Item):
     ) -> None:
         self.name = name
         self.defaults = defaults
+
+    def lint(self, lint_result: LintResult, context: LintContext) -> None:
+        super().lint(lint_result, context)
+        default_target = self.defaults.get("target") if self.defaults else None
+        if not default_target:
+            lint_result.add_warning(
+                f'Deployment "{self.name}" has no defaults.target set which is required for deployment auto-creation.',
+            )
+            return
--- a/tests/test_deployment.py
+++ b/tests/test_deployment.py
@@ -24,6 +24,11 @@ def test_minimal_valid_but_ineffective():
     assert deployment.name == "my-deployment"
     assert deployment.defaults is None
 
+    # it is acceptable, but we warn that it doesn't auto-create deployments
+    lint_result = config.lint()
+    assert lint_result.warning_count == 1
+    assert 'Deployment "my-deployment" has no defaults.target' in next(lint_result.warnings)["message"]
+
 
 def test_minimal_that_will_create_a_deployment():
     config = Config.parse(


examples/pipeline-with-automatic-deployment-creation.yaml
```diff
@@ -0,0 +1,69 @@
+# Defining deployments in the YAML is optional but allows Valohai platform to
+# initialize deployments automatically when a related pipeline is run.
+#
+# You may reference this deployment definition in the pipeline deployment
+# nodes like further down below.
+- deployment:
```
akx, line 6: I kind of want to call this `deployment-template` 🤔 because the other things here are pretty concrete and, like, _certainly_ there to choose from, but this'd only materialize into a deployment if a deployment node needs it?

Is that too verbose? 🤔

valohai_yaml/objs/deployment.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(
+        self,
+        *,
+        target: str,
+        allow_cors: Optional[bool] = None,
+        allow_configuration_snippets: Optional[bool] = None,
+    ) -> None:
+        self.target = target
+        self.allow_cors = allow_cors
+        self.allow_configuration_snippets = allow_configuration_snippets
```
akx, line 19: I would maybe not strictly specify the defaults type at all in valohai-yaml, to  give us a little more wiggle room to change things.

`defaults` could just be a `dict[str, Any]` bag, and we'd deal with it in the app.

`target` could be made soft-required by way of a lint warning in `Deployment`.

valohai_yaml/objs/deployment.py
```diff
@@ -0,0 +1,41 @@
...
+
+class DeploymentDefaults(Item):
+    """Represents the default values to use *if* matching deployment does not exist in the project context."""
+
+    def __init__(
+        self,
+        *,
+        target: str,
+        allow_cors: Optional[bool] = None,
+        allow_configuration_snippets: Optional[bool] = None,
```
akx, line 15: Btw, are these tristate (true/false/unspecified) on purpose?

valohai_yaml/objs/deployment.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class Deployment(Item):
+    """Represents a deployment in a project under where deployment versions and endpoints are housed."""
+
+    def __init__(
+        self,
+        *,
+        name: str,
+        defaults: Optional[DeploymentDefaults] = None,
```
akx, line 29: A `Deployment` without a `defaults` is pretty useless though, right?

valohai_yaml/objs/deployment.py
ruksi: I had it as a dict first but noticed that it was nicer to use with type hints during tests when it was a real class

I... can change it back if you want

how do you see it helping with wiggle room? at least with types you'd know you have wrong version of Valohai YAML in the context you are working in :laughing:

valohai_yaml/objs/deployment.py
ruksi: yes! does basically nothing except if you want to use it to document expected deployments the project needs

like I mentioned in the other comment, the `deployment` definition could be used as a baseline when creating deployments on the web UI so it might be helpful to suggest names for deployments, but that is not inside the current scope I'm working on

valohai_yaml/objs/deployment.py
ruksi: I changed it to a `dict` in either case!

valohai_yaml/objs/deployment.py
ruksi: this I left as-is, let me know if you want changes

ruksi:
massaged the implementation according to comments

- `defaults` are now a dict
- added a lint warning about missing `deployment.defaults.target`

akx (changes requested):
Tiny JSON Schema related change request still 😅

ruksi:
> Tiny JSON Schema related change request still 😅

scheme less strict now!

PR #174, opened 2025-09-02
Allow reusing task's execution from yaml config

dangquangdon:
Add property `reuse-executions` to task yaml definition


akx (changes requested):
Let's call this `reuse-children`.

akx (approved):

PR #177, opened 2025-10-27
Add support for non-local steps in PipelineConverter too

akx:
The YAML-to-API converter didn't have support for non-local steps...


teroyks (approved):
LGTM!
The test helps understand how the thing behaves. 👍

PR #180, opened 2025-12-09
🔢🎛️ Serialize parameter widget definitions

ruksi:
Refs https://github.com/valohai/roi/issues/9391

one might assume that `Parameter.get_data()` returns something that can be JSON serialized

but **not** when there is a widget definition

fix inside

### DIFF
--- a/tests/test_serialize_parameter.py
+++ b/tests/test_serialize_parameter.py
@@ -1,0 +1,40 @@
+import json
+
+from valohai_yaml.objs import Parameter
+from valohai_yaml.objs.parameter_widget import ParameterWidget
+
+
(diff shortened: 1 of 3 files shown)


akx (commented):
Good catch.

However: our tests that use `examples` are supposed to also test via roundtrip (see `config_fixture`) and it looks like `example1.yaml` has a widget. Why didn't that trigger this bug?

ruksi:
which would mean that the usage on ref `#9391` is wrong, not that `get_data` is wrong

will investigate further!

ruksi:
closing as the wrong approach; ref https://github.com/valohai/roi/pull/9408

Issue #145, opened 2024-03-05
☸️ Add support for runtime config preset in the environment property


teroyks:
Allow defining a `runtime_config` preset with the environment.
Both preset ID and slug should be supported.

Example (under `step` configuration):

```yaml
# with UUID
environment: g2.superduperlarge/f1bd21c8-ae31-451f-b75d-51cd0882aa1f

# with slug
environment: g2.superduperlarge/gpu-devices

# QOL: allow spaces for readability
environment: g2.superduperlarge / gpu-devices
```

teroyks:
Note: before adding any more preset-related features, the planned simplification of preset handling should be implemented!

PR #181, opened 2025-12-10
🏷️ Add `runtime-config-preset` property to `Step`

teroyks:
Refs #145 

Does not do anything with the property; added so that parsing can be implemented without creating a strict version dependency with Roi.

1. Add support for type. <- _done here_
2. Parse property in Roi. <- TBD next
3. Add value from YAML.


teroyks:
Included in #183 so closing this one to keep the number of review requests down.

PR #184, opened 2025-12-12
Become 0.52.0

### LINKED ISSUE #182: Release `valohai-yaml` with preset support in Step
Required as a Roi dependency to properly handle the extended `Step` objects.

teroyks:
New features in 0.52.0:

- Parse runtime config preset ID from YAML config (#183)

Resolves #182


ruksi (approved):
_let it be so_

PR #190, opened 2026-01-15
📝 Add description to pipeline edges in JSON schema

teroyks:
The edge definition content is not defined by the schema – add a bit of extra information to help the user (so they don’t have to dig in the docs for the content).

Pick the edge types from the actual definition so they stay up to date automatically (move the definition to a separate `types` module to avoid complicated import chains).

The description in the documentation:


teroyks:
Are the titles meaningful? Maybe they could also be improved…

akx (approved):
How descriptive!

PR #189, opened 2026-01-15
🔨 Generate Markdown documentation for the JSON schema

teroyks:
Resolves #179 

Run the following command to output a Markdown document of the JSON schema:

```shell
python -m valohai_yaml --output-schema-doc
```

***Note:*** There are _quite a few_ commits here, but I’d like to keep them: they might be useful later when wondering why some part of the code is the way it is (and give insight to how to improve on them).

***Another note:*** This is a MVP, so further improvements for legibility and organization can (and should) be done; this should already be usable for describing all the options for `valohai.yaml`, defined by the schema.

There are quite a few helper functions, especially doing the formatting. This is how the module is structured:

```mermaid
  graph TD
      Start[generate_schema_doc] --> PTL[parse_top_level_item_refs]
      Start --> PD[parse_definitions]
      PD --> PDF[parse_definition]
      Start --> FDC[format_doc_content]

      FDC ---> FRL[_format_ref_link]
      FDC --> FP[format_property]

      FP --> FPN[_format_prop_name]
      FP --> GTD[_get_type_definition]
      FP --> FP_recursive[format_property<br/>recursive]
      FP --> FLP[_format_list_property]
      FP --> FAP[_format_atomic_property]

      FLP --> GI[_get_indentation]
      FLP --> FLV[_format_list_value]

      FLV --> GI[_get_indentation]
      FLV --> GTD
      FLV ---> FRL[_format_ref_link]
      FLV --> FP2[format_property<br/>recursive]

      FAP ---> FRL[_format_ref_link]

      style Start fill:#e1f5ff
      style FP fill:#ffe1e1
      style FP_recursive fill:#ffe1e1
      style FP2 fill:#ffe1e1
```

This is the doc generated from the current schema (also included as a PDF for easier previewing):

[schema-doc.md](https://github.com/user-attachments/files/24642821/schema-doc.md)
[schema-doc.pdf](https://github.com/user-attachments/files/24642845/schema-doc.pdf)

PR #191, opened 2026-01-16
📝 Inline property title and description

teroyks:
Related to #189 

Improve document readability by formatting the title and description – properties meant to be human-readable – better, i.e as parts of the property presentation instead of as separate “sub-properties”.

Before:

After (dotted line added just for highlighting the updated area):


akx (approved):
Neater!

PR #197, opened 2026-01-30
Add possibility to override runtime_config_preset in pipeline node

teroyks:
Refs https://github.com/valohai/roi/issues/9452

- Add `runtime_config_preset` to overridable fields.
- Move handling of `runtime_config_preset` in step to parse_common_step_properties so it takes care of both normal step and override.

_Note:_ Requires backend support before the override can actually be used.


ruksi (approved):
new field to overrides

lgtm!

PR #199, opened 2026-02-03
✨ Add `deployment` edge type

### LINKED ISSUE #196: Implement deployment edges
Deployment edges exists in roi but not in valohai-yaml. It would be useful to be able to create nodes after a deployment node for example to check that the endpoint works as expected or to clean old endpoints. 

Passing either the deployment id or the version id via the edge would be great. 
```
    edges:
      - [Deploy.deployment.id, Cleanup.parameters.what_was_deployed]
      - [Deploy.deployment.version_id, Cleanup.parameters.version_id]
```

teroyks:
Resolves #196 

Deployment edge handling already exists in the backend; this just adds `deployment` as a valid edge type.


valohai_yaml/objs/pipelines/types.py
```diff
@@ -1,9 +1,10 @@
...
+    "file",
     "input",
+    "metadata",
     "output",
     "parameter",
-    "metadata",
-    "file",
-    "dependency",
-    "environment-variable",
 }
```
ruksi: sort + `deployment`

PR #206, opened 2026-03-24
Ensure all Items and Enums are re-exported

akx:
We were re-exporting an arbitrary selection of `objs` from the top-level package; this ensures all of them are, with a test to ensure it in the future.


ruksi (approved):
cleanup of topper-level exports

lgtm

and test to keep it so, good!

PR #208, opened 2026-03-30
Become 0.56.0

akx:
Follows up on #202.


akx:
Self-merging to get a release party on.

PR #209, opened 2026-04-10
Allow strings and booleans in variant-param-rule-item items

akx:
When #98 introduced task blueprints, we didn't think about parameters possibly being e.g. grid search over algorithm spec strings, etc., so the schema was overly strict.

### DIFF
--- a/valohai_yaml/schema_data.py
+++ b/valohai_yaml/schema_data.py
@@ -789,7 +789,7 @@ register(
             "distribution": {"type": "string"},
             "end": {"type": "number"},
             "integerify": {"type": "boolean"},
-            "items": {"items": {"type": "number"}, "type": "array"},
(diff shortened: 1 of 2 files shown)


valohai_yaml/schema_data.py
```diff
@@ -789,7 +789,7 @@ def _get_execution_and_task_node_props(type: str) -> dict[str, Any]:
             "distribution": {"type": "string"},
             "end": {"type": "number"},
             "integerify": {"type": "boolean"},
-            "items": {"items": {"type": "number"}, "type": "array"},
+            "items": {"items": {"type": ["number", "string", "boolean"]}, "type": "array"},
```
ruksi, line 792: strings and booleans are fine too!

ruksi (approved):
looks good to me

PR #213, opened 2026-05-29
☸️✍️ Allow `ephemeral-storage` settings on endpoint definitions

ruksi:
sometimes it can be useful to control [the ephemeral storage](https://kubernetes.io/docs/concepts/storage/ephemeral-storage/#requests-limits) per endpoint pod

- new optional `endpoint.resources.ephemeral-storage.min|max`
- added a few format comments to the main endpoint example as that is probably the first place people encounter these settings (schema info also describes them)

### DIFF
--- a/examples/endpoint-example.yaml
+++ b/examples/endpoint-example.yaml
@@ -24,8 +24,8 @@
         min: 0.1
         max: 1
       memory:
-        min: 50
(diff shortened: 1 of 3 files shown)


akx (approved):
Ephemeral!

PR #215, opened 2026-08-12
Skip linting of the step paramters if a node is referencing commit

SaltyPretzel303:
Linting such pipeline: 
```
- pipeline:
    name: commit_pipe
    nodes:
      - name: node_2
        type: execution
        step: some_step
        commit: <commit id>

    parameters:
    - name: main_param
      default: "value"
      targets: [node_2.parameters.main_param]
```

Will report an error as `some_step` step can't be found withing the steps in this .yaml.

Similarly how execution_node linting is skipped if commit is specified: https://github.com/valohai/valohai-yaml/blob/5f71bb7021557d5f4b64fe8e40fedb3ab482ec52/valohai_yaml/objs/pipelines/execution_node.py#L44
linting of a step parameters can also be skipped if a commit is specified.

### DIFF
--- a/examples/pipeline-with-param-referring-to-different-commit.yaml
+++ b/examples/pipeline-with-param-referring-to-different-commit.yaml
@@ -1,0 +1,15 @@
+- pipeline:
+    name: commit_pipe
+    edges: []
+    nodes:
+      - name: node_2
+        type: execution
+        step: some_step
+        commit: something-external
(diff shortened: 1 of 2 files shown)


akx (approved):
LGTM. Added the example from the PR message as a test case (tested it failed on `master`, works here).

PR #220, opened 2026-08-19
Put 'autorestart' inside 'runtime_config'

teroyks:
Move step data to `runtime_config`—where Roi reads execution-specific settings from.
(This is similar to existing handling of `no_output_timeout`.)


ruksi (approved):
ok, so it moves inside the runtime config

cool!

PR #222, opened 2026-08-24
☸️💽 Allow `shared-volumes` setting on endpoints

ruksi:
## 🚧 ${\color{yellow}Waiting For The App Side Of This To Get Merged}$ 🚧

allows defining Deployment Endpoint _shared volumes_, shaped:

```yaml
- endpoint:
    # ...
    shared-volumes:
      - pvc-name: my-shared-models
        mount-path: /models
        sub-path: predict-models
        read-only: true
```

where `sub-path` and `read-only` are optional


akx (approved):
Voluminous!

